PR #6, opened 2021-04-08
Move common scripts into testing repository

RedbackThomson:
This PR aims to set up the `test-infra` repository to support each service controller maintaining its own integration (e2e) testing code.

Tracking the wider project in https://github.com/aws-controllers-k8s/community/issues/735

Changes:
- Rebase Python scripts as found in `community`
- Replace `setup.cfg` with `setup.py` for pip compatibility
- Migrate `run-tests.sh` and accompanying resources from `community/scripts`
- Add documentation on usage (in `README.md`) and local development (in `/docs/setup.md`)


src/ack_test_infra/k8s/resource.py
```diff
@@ -0,0 +1,381 @@
...
+
+def create_k8s_namespace(namespace_name: str):
+    _api_client = _get_k8s_api_client()
+    return client.CoreV1Api(_api_client).create_namespace(
+        client.V1Namespace(name=namespace_name))
+
+
+def delete_k8s_namespace(namespace_name: str):
+    _api_client = _get_k8s_api_client()
+    return client.CoreV1Api(_api_client).delete_namespace(namespace_name)
```
jaypipes, line 146: How about placing these in a separate `namespace.py` module and calling them just `create` and `delete`?

Then, callers could do:

```python
from acktest.k8s import namespace

namespace.create(...)

namespace.delete(...)
```
...

jaypipes: We could take the same approach for secrets...

```python
from acktest.k8s import secret

secret.create_opaque(...)

secret.delete(...)

File: src/ack_test_infra/k8s/resource.py [hunk 2]
```diff
@@ -0,0 +1,381 @@
...
+    def _printable_namespace(self):
+        return self.namespace or "cluster"
+
+    def to_short_resource_string(self):
+        return f"{self._printable_namespace}:{self.name}"
+
+    def to_long_resource_string(self):
+        return f"{self.plural}.{self.version}.{self.group}/{self._printable_namespace}:{self.name}"
+
+def load_resource(service_name: str,
```
jaypipes, line 58: I'd love to add `create()` and `delete()` methods to the `CustomResourceReference` object above and have this function be called just `from_file` and make the calling sequence similar to this eventually:

```python
from acktest.k8s import resource

from test.e2e import SERVICE_NAME
...
my_res = resource.from_file(SERVICE_NAME, "path/to/resource.yaml", replacements)
...
my_res.create()
...
my_res.delete()
```

We could then make all the functions in this file private and hide the implementation details as much as possible from the callers.

src/ack_test_infra/resources.py
```diff
@@ -25,10 +25,14 @@
 
 from .aws import identity
 
-PLACEHOLDER_VALUES = {
-    "AWS_ACCOUNT_ID": identity.get_account_id(),
-    "AWS_REGION": identity.get_region(),
-}
+@property
+def PLACEHOLDER_VALUES():
```
jaypipes, line 29: Any reason to make this into a property?

setup.py
```diff
@@ -0,0 +1,29 @@
+from setuptools import setup, find_packages
```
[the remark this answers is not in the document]
RedbackThomson: Yeah `pip install` does not support `setup.cfg`. There was also `pyproject.toml`, which is outlined in PEP 517/518, but it's not yet well supported.

src/ack_test_infra/k8s/resource.py
RedbackThomson: This is a more object oriented, fashion? Do we want to do this with each of the types?

src/ack_test_infra/resources.py
RedbackThomson: The global values was (attempting to be) initialised when importing this script every time. It was causing errors when loading the package outside the scope of the PyTests.

scripts/build-run-test-dockerfile.sh
```diff
@@ -0,0 +1,57 @@
...
+echo "Running e2e test container $TEST_DOCKER_SHA"
+# Ensure it can connect to KIND cluster on host device by running on host 
+# network. 
+# Pass AWS credentials and kubeconfig through to Dockerfile.
+docker run --rm -t \
+    --network="host" \
+    -v $KUBECONFIG_LOCATION:/root/.kube/config:z \
+    -v $HOME/.aws/credentials:/root/.aws/credentials:z \
+    -v $THIS_DIR:/root/tests:z \
+    -e SERVICE_CONTROLLER_E2E_TEST_PATH="." \
```
jaypipes, line 51: The arg in the Dockerfile is CONTROLLER_E2E_PATH, but the arg in the run-tests.sh script is SERVICE_CONTROLLER_E2E_TEST_PATH. Might want to make them consistent..

setup.py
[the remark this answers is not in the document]
jaypipes: > Yeah `pip install` does not support `setup.cfg`. There was also `pyproject.toml`, which is outlined in PEP 517/518, but it's not yet well supported.

Huh, well TIL... :)

jaypipes (approved):
:+1:

PR #9, opened 2021-04-17
Add helper function to check resource deletion

surajkota:
### Testing
[aws-controllers-k8s/sagemaker-controller#14](https://github.com/aws-controllers-k8s/sagemaker-controller/pull/14)


src/acktest/k8s/resource.py
```diff
@@ -151,6 +151,21 @@ def patch_custom_resource(
     return _api.patch_namespaced_custom_object(
         reference.group, reference.version, reference.namespace, reference.plural, reference.name, custom_resource)
 
+def assert_delete_resource_successful(
```
jaypipes, line 154: This method isn't asserting anything.

PR #12, opened 2021-04-23
Split helm charts and add service account roles

RedbackThomson:
- Splits out the Prow Helm charts based on the [`eks-distro-build-tooling`](https://github.com/aws/eks-distro-build-tooling/tree/main/helm-charts/stable) example. 
- CDK now creates ServiceAccounts (and registers them with IAM roles) with permissions for accessing S3
  - Passing service account names through Helm
- Fix: Create namespaces as CDK8s chart after creating cluster, before installing anything else. As a precaution for race conditions messing with different manifests that assume existence.


prow/data-plane/values.yaml
```diff
@@ -0,0 +1,37 @@
...
+  serviceAccount:
+    name: ''
+
+prowControllerManager:
+  serviceAccount:
+    name: ''
+
+sinker:
+  serviceAccount:
+    name: ''
```
vijtrip2, line 25: Will this empty make SA name default to 'sinker' in the manifest file?

vijtrip2 (approved):
LGTM!

prow/data-plane/values.yaml
RedbackThomson: Yeah in all of the service account name fields we default every name to the name of the deployment - that's the standard for Prow.

PR #15, opened 2021-04-27
Add support for DNS and ALB controllers

RedbackThomson:
- Adds the `aws-load-balancer-controller` to create an ALB for Prow ingress
- Adds `external-dns` to support automatically updating the ALB ingress controller


vijtrip2 (approved):
/lgtm

PR #17, opened 2021-05-03
Adding the release-controller postsubmit prowjob.

vijtrip2:
Here I am ,
* adding a release-controller prow-job which triggers on semver tag creation on apigatewayv2-controller repository.
Sample [prow-job runs](https://prow.ack.aws.dev/?repo=vijtrip2%2Fapigatewayv2-controller&type=postsubmit&state=success) and [logs](https://prow.ack.aws.dev/log?job=apigatewayv2-post-submit&id=1389094606846562304) 
* adding support for publishing helm chart when changes are pushed in stable branch.
* Also updating the prow controller images to work with latest configuration.


cd/prow/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+		git \
+		unzip \
+		openssl \
+    && curl -sL https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip -o awscliv2.zip \
+    && unzip awscliv2.zip \
+    && aws/install \
+    && curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 \
+    && chmod +x get_helm.sh \
+    && ./get_helm.sh
+ENTRYPOINT ["/bin/bash", "-c", "echo Hello from ACK Prow CD container."]
```
RedbackThomson, line 14: Probably not necessary? :P

cd/prow/Dockerfile
```diff
@@ -0,0 +1,14 @@
+# Container image for running ACK CD prow jobs.
+FROM quay.io/containers/buildah
```
RedbackThomson, line 2: Do they have an LTS build tag we can cache?

cd/prow/Dockerfile
```diff
@@ -0,0 +1,14 @@
+# Container image for running ACK CD prow jobs.
```
RedbackThomson, line 1: So ideally I'd like to take this container and re-use it for integration tests, since it seems to have all the necessary tools (buildah, git, awscli). I think this might serve as the base image for the integration test Dockerfile, as part of a multi-stage build.

infra/lib/prow-service-accounts.ts
```diff
@@ -50,6 +50,34 @@ export class ProwServiceAccounts extends cdk.Construct {
...
+        "ecr-public:UploadLayerPart",
+        "ecr-public:CompleteLayerUpload",
+        "ecr-public:InitiateLayerUpload",
+        "ecr-public:PutImage",
+        "ecr-public:ListTagsForResource",
+        "ecr-public:PutRegistryCatalogData",
+        "ecr-public:BatchCheckLayerAvailability"
+      ],
+      resources: [
+        `arn:${props.stackPartition}:ecr-public::${process.env.CDK_DEFAULT_ACCOUNT}:registry/*`,
```
RedbackThomson, line 69: I've read that this environment variable does not always exist. Elsewhere in the CDK I pass down the `account` as part of props. https://github.com/aws-controllers-k8s/test-infra/blob/8c107c357e86233ac4ccf1a8a56d34eaa793dc69/infra/lib/test-ci-stack.ts#L20

prow/config/templates/plugins-ConfigMap.yaml
```diff
@@ -18,19 +18,25 @@ metadata:
   name: plugins
 data:
   plugins.yaml: |
+    triggers:
+    - repos:
+      - {{ .Values.github.organisation }}
+      only_org_members: true
```
RedbackThomson, line 24: This is a key-value pair as part of the repos list?

prow/config/templates/plugins-ConfigMap.yaml
```diff
@@ -18,19 +18,25 @@ metadata:
...
+        - dog
+        - help
+        - heart
+        - hold
+        - label
+        - lgtm
+        - trigger
+        - verify-owners
+        - wip
+        - yuks
```
RedbackThomson, line 42: Could we crawl through these and pull out the silly ones? dog, cat and yuks for example.

prow/jobs/jobs.yaml
```diff
@@ -6,4 +6,26 @@ periodics:
     serviceAccountName: pre-submit-service-account
     containers:
     - image: alpine
-      command: ["/bin/date"]
\ No newline at end of file
+      command: ["/bin/date"]
+
+postsubmits:
+  #TODO(vijat@): Add support for all service controllers
+  #TODO(vijat@): Find a programmatic way to generate postsubmit jobs for all the service controllers
```
RedbackThomson, line 13: Ideally I'd split out `jobs.yaml` into multiple yaml files that are all concatenated. I believe there is a way, as I've seen other projects do it. Possibly as part of that we could have a codegen for each service to produce its own set of job files, per.

prow/jobs/jobs.yaml
```diff
@@ -6,4 +6,26 @@ periodics:
...
+  aws-controllers-k8s/apigatewayv2-controller:
```
RedbackThomson, line 14: Can we turn this into a preset and re-use it more easily that way?

prow/jobs/jobs.yaml
```diff
@@ -6,4 +6,26 @@ periodics:
...
+      base_ref: main
+      workdir: true
+    spec:
+      serviceAccountName: post-submit-service-account
+      containers:
+        - image: errorbyte/ack-prow-cd #TODO(vijat@): find the correct place for hosting this image.
+          securityContext:
+            privileged: true
+          command: ["/bin/bash"]
+          args: ["-c", "cd cd/scripts && ./release-controller.sh"]
```
RedbackThomson, line 29: Should this be the default entrypoint? I worry that if we make a change to the image we will have to remember to come back into here and modify all of the commands

prow/config/templates/plugins-ConfigMap.yaml
A-Hilaly: Agreed they are not the most useful plugins but they are fun 😂  - i vote for keeping them for now

cd/scripts/release-controller.sh
```diff
@@ -0,0 +1,121 @@
...
+
+cd "$WORKSPACE_DIR"
+
+if [[ $QUIET = "false" ]]; then
+    echo "building '$AWS_SERVICE' controller docker image with tag: ${AWS_SERVICE_DOCKER_IMG}"
+    echo " git commit: $SERVICE_CONTROLLER_GIT_COMMIT"
+fi
+
+# build controller image
+buildah bud \
```
A-Hilaly, line 80: Ohh TIL!

cd/prow/Dockerfile
vijtrip2: I will use one of the standard release rather than latest.

cd/prow/Dockerfile
vijtrip2: I will update the description.

prow/config/templates/plugins-ConfigMap.yaml
vijtrip2: No, it exists alongside the repos list inside Trigger struct.
and "triggers" is list of Trigger struct.
https://sourcegraph.com/github.com/kubernetes/test-infra/-/blob/prow/plugins/config.go#L380:1

RedbackThomson: Ugh reading YAML can be a pain in the butt sometimes. I see it now, though.

prow/jobs/jobs.yaml
vijtrip2: We do not use any properties of podspec supported by presets. So I do not see how we can use presets here.
https://github.com/kubernetes/test-infra/blob/3afb608d28630b99e49e09dd101a96c201268739/prow/config/jobs.go#L35

Presets support Label, Env, Volume and VolumeMount

infra/lib/prow-service-accounts.ts
vijtrip2: Done.

prow/jobs/jobs.yaml
vijtrip2: Looking into possible solutions for this.

RedbackThomson: Stretch goal for now. Not necessary as part of this PR, ofc.

prow/jobs/jobs.yaml
```diff
@@ -6,4 +6,272 @@ periodics:
...
+      repo: test-infra
+      base_ref: main
+      workdir: true
+    spec:
+      serviceAccountName: post-submit-service-account
+      containers:
+        - image: errorbyte/ack-prow-cd #TODO(vijat@): find the correct place for hosting this image.
+          securityContext:
+            privileged: true
+          command: ["/bin/bash", "-c", "cd cd/scripts && ./release-controller.sh"]
```
RedbackThomson, line 27: Is this not the default entrypoint command for the containers? Why do we need to keep using this?

prow/jobs/jobs.yaml
```diff
@@ -6,4 +6,272 @@ periodics:
...
+    - org: aws-controllers-k8s
+      repo: test-infra
+      base_ref: main
+      workdir: true
+    spec:
+      serviceAccountName: post-submit-service-account
+      containers:
+        - image: errorbyte/ack-prow-cd #TODO(vijat@): find the correct place for hosting this image.
+          securityContext:
+            privileged: true
```
RedbackThomson, line 26: This scares me :/ Is this because of `buildah` failing otherwise?

prow/jobs/jobs.yaml
vijtrip2: Prow mandates that podspec mentions the command and not use container image's default entrypoint.

prow/jobs/jobs.yaml
vijtrip2: Correct.

cd/scripts/release-controller.sh
```diff
@@ -0,0 +1,152 @@
...
+
+Publishes the controller image and helm chart for an ACK service controller as part of prowjob
+when a service controller repository is tagged with a semver format '^v\d+\.\d+\.\d+$'
+See: https://github.com/aws-controllers-k8s/test-infra/prow/jobs/jobs.yaml for prowjob configuration.
+
+Environment variables:
+  REPO_NAME:                Name of the service controller repository. Ex: apigatewayv2-controller
+                            This variable is injected into the pod by Prow.
+  PULL_BASE_REF:            The value of tag on service controller repository that triggered the
+                            postsubmit prowjob. The value will be in the format '^v\d+\.\d+\.\d+$'.
```
jaypipes, line 17: According to the below, the value might also be "stable"?

cd/scripts/release-controller.sh
```diff
@@ -0,0 +1,152 @@
...
+SERVICE_CONTROLLER_DIR="$WORKSPACE_DIR/$AWS_SERVICE-controller"
+
+# Check all the dependencies are present in container.
+source "$SCRIPTS_DIR"/lib/common.sh
+check_is_installed buildah
+check_is_installed aws
+check_is_installed helm
+check_is_installed git
+
+if [[ $PULL_BASE_REF = stable ]]; then
```
jaypipes, line 51: The Git branch will be `$MAJOR_VERSION-stable` I believe.

jaypipes (commented):
>     * find the repository for hosting 'ack-prow-cd' image that runs as container for release prow-job.

Instead of the `errorbyte/` repo that is in there now? What is that repo?

vijtrip2:
> > ```
> > * find the repository for hosting 'ack-prow-cd' image that runs as container for release prow-job.
> > ```
> 
> Instead of the `errorbyte/` repo that is in there now? What is that repo?

It was a personal repo as placeholder. I am now using `public.ecr.aws/aws-controllers-k8s/prow:prow-job-base-v0.0.1` as the image. LMK if you have any suggestions.

cd/scripts/release-controller.sh
vijtrip2: Yes, Thanks! Updated.

cd/scripts/release-controller.sh
vijtrip2: This will be the name of branch, which is 'stable' as per the doc content in community repo. 
> "In the stable Git branch of the ElastiCache service controller's source repository,"

From this variable I create VERSION variable which is $MAJOR_VERSION-stable

prow/jobs/jobs.yaml
vijtrip2: Added autogeneration of jobs using jinja template.

Makefile
```diff
@@ -0,0 +1,12 @@
+SHELL := /bin/bash # Use bash syntax
```
RedbackThomson, line 1: Makefile instead of bash script?

RedbackThomson: Makefile can call a bash script instead?

prow/jobs/jinja/postsubmits/controller_release.jinja2
```diff
@@ -0,0 +1,21 @@
+{% for service in aws_services  %}
```
RedbackThomson, line 1: Such a good idea!

prow/jobs/jobs_config.yaml
```diff
@@ -0,0 +1,23 @@
+#This config is used to generate 'jobs.yaml' file from 'jobs_factory.py' using jinja2 templates.
+
+# specify which kind of jobs are present
```
RedbackThomson, line 3: Could you clarify what it means for it to be "present"? I imagine in the future teams would want to run custom integration/canary/soak tests that we would need to expand out.

requirements.txt
```diff
@@ -2,3 +2,4 @@ boto3==1.16.40
 kubernetes==12.0.1
 PyYAML==5.4
 pytest-xdist==2.2.0
+Jinja2==2.11.3
```
RedbackThomson, line 5: This is not the place for this. These requirements are for the `acktest` source package that all repositories import as part of their integration tests. I would add a new `requirements.txt` in `prow/jobs/` alongside the Python files, there.

RedbackThomson: Although I think moving the prow jobs to a new repository altogether would be a better bet. But for now here is fine.

vijtrip2: I will cleanup and separate the `requirements.txt` for now.

prow/jobs/jobs_config.yaml
vijtrip2: Oh, it is just which kind of jobs out of periodics/postsubmits/presubmits are present.
I added this so that the auto-generation script does not emit an empty "presubmits:" in jobs.yaml, but I can handle it better in the script logic itself.

I will update this part.

Makefile
vijtrip2: I believe makefiles are just more developer friendly.
make <something> sounds better to me than running a bash script. (Just a personal opinion.)

In future there can be more make targets and IMO these 3-4 commands are not worthy of separate bash script.

RedbackThomson (approved):

PR #34, opened 2021-05-26
Remove insecure default test role

RedbackThomson:
As per the AppSec review, the default test role will create resources and have access to any secrets in the infrastructure AWS account. All service teams should onboard to their own Prow resource accounts, as tracked in https://github.com/aws-controllers-k8s/community/issues/797.


jaypipes (approved):
++

PR #35, opened 2021-05-26
Update cluster size and pytest parallel threads

RedbackThomson:
Updates the size of the cluster from the default 2x m5.large (4 vCPUs total) to 2x m5.8xlarge (64 vCPUs total).

Add a `PYTEST_NUM_THREADS` environment variable to allow fine-grained control over the number of threads/processes pytest-xdist uses.


jaypipes (approved):
tres bien. :)

PR #37, opened 2021-05-27
Add support for loading incluster kubeconfig for soak tests.

vijtrip2:
Adding support for loading incluster kubeconfig for soak tests.

This helps in running soak-tests using job/pod from the same kubernetes cluster where the ack controller is running.

Tested by running soak tests in a local cluster.

### DIFF
--- a/src/acktest/k8s/resource.py
+++ b/src/acktest/k8s/resource.py
@@ -16,6 +16,8 @@ CustomResource APIs.
 
 import logging
 import base64
+import os
+import distutils.util as util
 from pathlib import Path
 from time import sleep
 from typing import Dict, Optional, Union
@@ -109,10 +111,14 @@ def load_and_create_resource(resource_directory: Path,
     resource = create_resource(reference, spec)
     return reference, spec, resource
 
+
 def _get_k8s_api_client() -> ApiClient:
     # Create new client everytime to avoid token refresh issues
     # https://github.com/kubernetes-client/python/issues/741
     # https://github.com/kubernetes-client/python-base/issues/125
+    if util.strtobool(os.environ.get('LOAD_IN_CLUSTER_KUBECONFIG', 'false')):
+        config.load_incluster_config()
+        return ApiClient()
     return config.new_client_from_config()
 
 


src/acktest/k8s/resource.py
```diff
@@ -109,10 +110,14 @@ def load_and_create_resource(resource_directory: Path,
     resource = create_resource(reference, spec)
     return reference, spec, resource
 
+
 def _get_k8s_api_client() -> ApiClient:
     # Create new client everytime to avoid token refresh issues
     # https://github.com/kubernetes-client/python/issues/741
     # https://github.com/kubernetes-client/python-base/issues/125
+    if os.environ.get('LOAD_IN_CLUSTER_CONFIG') is not None and os.environ.get('LOAD_IN_CLUSTER_CONFIG').lower() == 'true':
```
RedbackThomson, line 118: This conditional can be greatly simplified to 
```suggestion
if bool(distutils.util.strtobool(os.environ.get('LOAD_IN_CLUSTER_CONFIG', "false")): 
```

RedbackThomson (commented):
Currently for e2e tests we run all pytest inside a Docker container running inside the Prow pod. Will you be running this container directly within a pod in the cluster? (This would need to be achieved using a `Job` type, I assume)

vijtrip2:
> Currently for e2e tests we run all pytest inside a Docker container running inside the Prow pod. Will you be running this container directly within a pod in the cluster? (This would need to be achieved using a `Job` type, I assume)

Yes, Exactly. Soak tests will run as `Job` resource and not inside Docker, but directly in a container.

src/acktest/k8s/resource.py
vijtrip2: Thanks! I will test it out and then update the CR soon.

vijtrip2: > This conditional can be greatly simplified to

Done.

src/acktest/k8s/resource.py
```diff
@@ -109,10 +111,14 @@ def load_and_create_resource(resource_directory: Path,
     resource = create_resource(reference, spec)
     return reference, spec, resource
 
+
 def _get_k8s_api_client() -> ApiClient:
     # Create new client everytime to avoid token refresh issues
     # https://github.com/kubernetes-client/python/issues/741
     # https://github.com/kubernetes-client/python-base/issues/125
+    if util.strtobool(os.environ.get('LOAD_IN_CLUSTER_CONFIG', 'false')):
```
A-Hilaly, line 119: very small nit: s/`LOAD_IN_CLUSTER_CONFIG`/`LOAD_IN_CLUSTER_KUBECONFIG`/

A-Hilaly (approved):
👍

A-Hilaly (approved):
👍

PR #40, opened 2021-05-28
Adding initial instruction for running local soak tests.

vijtrip2:
* Adding initial instruction for running local soak tests.
* Add the ACK grafana dashboard template that service team's can import.

### DIFF
--- a/soak/Dockerfile
+++ b/soak/Dockerfile
@@ -1,0 +1,25 @@
+FROM python:3.8-alpine
+
+# Persist build arguments into environment variables
+ARG AWS_SERVICE
+ARG E2E_GIT_REF=main
+ARG CONTROLLER_E2E_PATH=./${AWS_SERVICE}-controller/test/e2e
+ENV AWS_SERVICE ${AWS_SERVICE}
+ENV PYTHONPATH ${CONTROLLER_E2E_PATH}
+ENV CONTROLLER_E2E_PATH ${CONTROLLER_E2E_PATH}
+
+WORKDIR /soak
+# Install dependencies for soak test environment
+RUN apk add --no-cache git bash gcc libc-dev
+
+# Copy the script to run soak tests.
+COPY run_soak_test.sh .
+RUN chmod +x run_soak_test.sh
+
+# Checkout the controller repository where e2e tests are present.
+# Soak test run consists of multiple runs of these e2e tests.
+RUN git clone https://github.com/aws-controllers-k8s/${AWS_SERVICE}-controller.git -b ${E2E_GIT_REF} --depth 1
+RUN cd ${AWS_SERVICE}-controller/test/e2e \
+    && pip install -r requirements.txt
+
+ENTRYPOINT ["bash", "-c", "./run_soak_test.sh"]
--- a/soak/README.md
+++ b/soak/README.md
@@ -1,0 +1,321 @@
+## Introduction
+
+ACK team provides the soak testing framework where service teams can execute long running soak tests for testing the
+performance and stability of the service controllers. Running soak tests is also a requirement for cutting the stable
+release for the service controller. Find more details on the release process [here](https://github.com/aws-controllers-k8s/community/blob/main/docs/contents/releases.md).
+
+Under the hood, the soak test introduces load by repeatedly running the end-to-end tests for the service controller. End
+to end tests are present in service-controller github repository. i.e. "https://github.com/aws-controllers-k8s/<service-name>-controller".
+
+If you wish to run soak tests for your service controller locally, see the "How To Run Soak Test Locally" section below.
+
+## Running Soak Tests Using ACK Prow Jobs
+
+TBD
+
+
+## How To Run Soak Test Locally
+
+### Prerequisites
+* A tool for building OCI images (docker, buildah etc..)
+* A Kubernetes cluster to run soak tests
+* helm 
+* kubectl
+* curl
+* yq
+
+### Step 0 (Declare the ACK soak test configuration)
+
+* Setup soak test directory
+    ```bash
+    mkdir -p ~/.ack/soak && cd ~/.ack/soak
+    ```
+
+* Create a file name `config` and copy the following configuration into that file, substituting "< placeholders >" with
+actual values.
+
+    ```bash
+    # NOTE: Substitute all the <placeholders> with actual values.
+    
+    # A handy alias to go to root of ACK soak test directory
+    alias go-to-soak='cd ~/.ack/soak'
+    
+    ### SERVICE CONTROLLER CONFIGURATION ###
+    
+    # AWS Service name of the controller under test. Ex: apigatewayv2
+    export SERVICE_NAME=<aws-service-name>
+    
+    # IAM Role Arn which will provide privileges to service account running controller pod
+    # This role will also provide access to the soak-test-runner for validating e2e test results
+    export IAM_ROLE_ARN_FOR_IRSA=<"IAM Role ARN for providing AWS access to ack controller's service account">
+
+    export HELM_EXPERIMENTAL_OCI=1
+    
+    # Semver version of the controller under test. Ex: v0.0.2
+    export CONTROLLER_VERSION=<semver>
+    export CONTROLLER_CHART_URL=public.ecr.aws/aws-controllers-k8s/chart:$SERVICE_NAME-$CONTROLLER_VERSION
+    
+    # AWS Region for ACK service controller
+    export CONTROLLER_AWS_REGION=us-west-2
+  
+    # AWS Account Id for ACK service controller
+    export CONTROLLER_AWS_ACCOUNT_ID=<aws-account-id>
+    
+    # Release name of controller helm chart.
+    export CONTROLLER_CHART_RELEASE_NAME=soak-test
+        
+    # Evaluation string for yq tool for updating helm chart's values.yaml file
+    export SERVICE_ACCOUNT_ANNOTATION_EVAL=".serviceAccount.annotations.\"eks.amazonaws.com/role-arn\" = \"$IAM_ROLE_ARN_FOR_IRSA\""
+    export METRIC_SERVICE_CREATE_EVAL=".metrics.service.create = true"
+    export METRIC_SERVICE_TYPE_EVAL=".metrics.service.type = \"ClusterIP\""
+    export AWS_REGION_EVAL=".aws.region = \"$CONTROLLER_AWS_REGION\""
+    export AWS_ACCOUNT_ID_EVAL=".aws.account_id = \"$CONTROLLER_AWS_ACCOUNT_ID\""
+    
+    ### PROMETHEUS, GRAFANA CONFIGURATION ###
+    
+    # Release name of kube-prometheus helm chart.
+    export PROM_CHART_RELEASE_NAME=kube-prom
+    
+    # Local port to access Prometheus dashbaord
+    export LOCAL_PROMETHEUS_PORT=9090
+    
+    # Local port to access Prometheus dashbaord
+    export LOCAL_GRAFANA_PORT=3000
+    
+    ### SOAK TEST RUNNER CONFIGURATION ###
+    
+    # Image repository where soak test runner image will be stored. (This can be your personal repository.) Ex: 'my-ack-repo/soak-test'
+    # Make sure this is a public repository.
+    export SOAK_IMAGE_REPO=<repository-to-store-soak-test-runner-image>
+    
+    # Image tag for soak-test-runner image.
+    export SOAK_IMAGE_TAG=0.0.1
+    
+    # Release name of soak-test-runner helm chart.
+    export SOAK_RUNNER_CHART_RELEASE_NAME=soak-test-runner
+    
+    # Total test duration is calculated as sum of TEST_DURATION_MINUTES, TEST_DURATION_HOURS and TEST_DURATION_DAYS after converting them in minutes.
+    # Override following variables accordingly to set your soak test duration. Default value: 24 hrs.
+    export TEST_DURATION_DAYS=1
+    export TEST_DURATION_HOURS=0
+    export TEST_DURATION_MINUTES=0
+    export NET_SOAK_TEST_DURATION_MINUTES=$(($TEST_DURATION_MINUTES + $TEST_DURATION_HOURS*60 + $TEST_DURATION_DAYS*24*60))
+    ``` 
+
+* Initialize the shell with ACK soak test configuration
+
+    ```
+    source ~/.ack/soak/config
+    ```
+
+Validation:
+* Running `echo $CONTROLLER_CHART_URL` should return non-empty result.
+
+### Step 1 (Install the ACK service controller using helm)
+
+In this step we install ACK service controller which will be soak tested using the soak-test runner created in step 4.
+
+You can install an ACK controller following the documentation [here](https://aws-controllers-k8s.github.io/community/user-docs/install/)
+but following commands are also TL;DR for the installation documentation.
+
+You can use either (a) or (b) method below.
+
+* Pull the helm chart either from (a) or (b)
+
+    a) from public repo
+
+    ```bash
+    go-to-soak \
+    && helm chart pull $CONTROLLER_CHART_URL \
+    && mkdir controller-helm \
+    && cd controller-helm \
+    && helm chart export $CONTROLLER_CHART_URL
+    && cd ack-$SERVICE_NAME-controller
+    ```
+    
+    b) from source repository
+    
+    ```bash
+    go-to-soak \
+    && git clone https://github.com/aws-controllers-k8s/$SERVICE_NAME-controller.git -b main --depth 1 \
+    && cd $SERVICE_NAME-controller/helm
+    ```
+
+* Insert $IAM_ROLE_ARN_FOR_IRSA into values.yaml
+    ```bash
+    yq e $SERVICE_ACCOUNT_ANNOTATION_EVAL -i values.yaml \
+    && yq e $METRIC_SERVICE_CREATE_EVAL -i values.yaml \
+    && yq e $METRIC_SERVICE_TYPE_EVAL -i values.yaml \
+    && yq e $AWS_REGION_EVAL -i values.yaml \
+    && yq e $AWS_ACCOUNT_ID_EVAL -i values.yaml
+    ```
+
+* Run
+    ```bash
+    helm install $CONTROLLER_CHART_RELEASE_NAME .
+    ```
+
+NOTE
+* IRSA setup is must for executing soak tests on ACK service controllers. See [installation docs](https://aws-controllers-k8s.github.io/community/user-docs/install/)
+ for how to use IRSA for ACK service controller pod.
+
+Validation:
+* After successful execution of above commands, ack-service-controller will be running in your cluster, exposing ACK 
+metrics through a K8s service endpoint.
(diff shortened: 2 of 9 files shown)


soak/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+RUN pip install -r requirements.txt
+
+# Copy the script to run soak tests.
+COPY soak/run_soak_test.sh .
+RUN chmod +x run_soak_test.sh
+
+# Checkout the controller repository where e2e tests are present.
+# Soak test run consists of multiple runs of these e2e tests.
+RUN git clone https://github.com/aws-controllers-k8s/${AWS_SERVICE}-controller.git
+RUN cd ${AWS_SERVICE}-controller && git checkout ${E2E_GIT_REF}
```
RedbackThomson, line 26: Git clone seems to offer the ability to choose the branch ? https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---branchltnamegt

Can we also do a clone depth of 0?

soak/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+WORKDIR /
```
RedbackThomson, line 28: Not necessary as specified on L11

soak/README.md
```diff
@@ -0,0 +1,202 @@
+## INTRODUCTION
```
RedbackThomson, line 1: Titles should be proper cased?

soak/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+ACK team provides the soak testing framework for service teams where service teams can execute long running
```
RedbackThomson, line 3: nit: Remove the `for service teams`

soak/README.md
```diff
@@ -0,0 +1,202 @@
...
+soak tests for testing the performance and stability of the service controllers.
```
RedbackThomson, line 4: To add: Requirement for stable version release

soak/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+Under the hood, these soak tests consist of multiple end to end test runs for the service controller. End to end tests are present in service-controller github repository.
```
RedbackThomson, line 6: nit: "consist of multiple end-to-end tests" I think we're implying here that "introduces load by repeatedly running the end-to-end tests"

soak/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+TODO
+```
+* Convert following 9 steps into a single executable script with default ack soak-test params. 
+* Service teams will have the freedom to set parameters and customizing their soak-test runs.
+```
+
+### Prerequisites
+* A tool for building OCI images (docker, buildah etc..)
+* A kubernetes cluster to run soak tests
```
RedbackThomson, line 26:
```suggestion
* A Kubernetes cluster to run soak tests
```

soak/README.md
```diff
@@ -0,0 +1,202 @@
...
+* helm 
+* kubectl
+
+
+### STEP 1 (Install the ACK service controller using helm)
+
+In this step we install ack service controller which will be soak tested using the soak-test runner created in step 4.
```
RedbackThomson, line 33: nit: `s/ack/ACK/g`

soak/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+You can install an ACK controller following the documentation [here](https://aws-controllers-k8s.github.io/community/user-docs/install/)
+but following commands are also TL;DR for the installation documentation.
+
+You can use either (a) or (b) method below.
+
+a) Pull the chart from public repo
```
RedbackThomson, line 40: This option should not be valid because soak tests are run before we release into the public chart repo. I'd only recommend (b)

soak/README.md
```diff
@@ -0,0 +1,202 @@
...
+* RUN `helm chart pull $CONTROLLER_CHART_URL`
+* RUN `helm chart install $CONTROLLER_CHART_RELEASE_NAME $CONTROLLER_CHART_URL` 
+
+b) Use the chart from source repository
+
+* RUN `export SERVICE_NAME=<aws-service-name>` Ex: 'apigatewayv2'
+* checkout `<aws-service-name>-controller` github repository
+* RUN `cd helm`
+* RUN `export CONTROLLER_CHART_RELEASE_NAME=<controller-chart-release-name>` Ex: soak-test
+* RUN `helm install $CONTROLLER_CHART_RELEASE_NAME .` Note: Set custom values during helm install
```
RedbackThomson, line 55: Can this just be concatenated into a single runnable bash block? I'd much rather just copy and paste something than read bullet points

RedbackThomson: This should also include instructions to set `service.create=true`?

soak/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+NOTE
+* Use "--set serviceAccount.Annotations.eks.amazonaws.com/role-arn=<IAM_ROLE_ARN_FOR_IRSA>" during helm install.
```
RedbackThomson, line 58: I would include this as mandatory in the above block - temporary credentials cannot be used because they would expire within 24hrs.

soak/README.md
```diff
@@ -0,0 +1,202 @@
...
+       * Now you can view Prometheus metrics at `http://127.0.0.1:9090/`
+    * GRAFANA
+        * Run `kubectl port-forward service/$PROM_CHART_RELEASE_NAME-grafana 3000:80`
+        * Now you can view Grafana dashboards at `http://127.0.0.1:3000/`
+        * Default Username for Grafana login is `admin` and default password is `prom-operator`
+        * Note: You can also see these username and password in secret `$PROM_CHART_RELEASE_NAME-grafana`
+* After logging into Grafana dashboard, to view resource metrics of controller-pod,
+    * Go to `http://127.0.0.1:3000/dashboards`
+    * Under `Manage` tab find `Kubernetes/Compute Resources/Pod`
+    * And load the metrics for service-controller pod from the pod selection.
```
RedbackThomson, line 90: We are going to need to automate this process for the soak tests? Is it possible to just provide the script to automate it here?

soak/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+NOTE: You can also choose to update Prometheus and Grafana services to NodePort or LoadBalancer Type service, if you wish
+to access them not through port-forward.
```
RedbackThomson, line 93: nit: `to access them through a public-facing ELB`

soak/helm/ack-soak-test/values.yaml
```diff
@@ -0,0 +1,13 @@
+# Default values for ack-soak-test.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+serviceName: "" # provide during helm install command
```
RedbackThomson, line 5: I initially understand this as K8s idea of "service" rather than the AWS service name. Perhaps this could be clarified as "awsService"?

soak/helm/ack-soak-test/values.yaml
```diff
@@ -0,0 +1,13 @@
...
+testJobParallelism: 1
+testJobCompletions: 1
+awsRegion: "us-west-2"
+soak:
+  imageRepo: "" # provide during helm install command
+  imageTag: "" # provide during helm install command
+  startTimeEpochSeconds: "" # provide during helm install command
+  durationMinutes: 90
```
RedbackThomson, line 13: Soak tests will need to be much longer than 90 minutes - default to 24 hours?

soak/run_soak_test.sh
```diff
@@ -0,0 +1,44 @@
...
+
+echo "[INFO] Starting the soak test run."
+echo "[INFO] START_TIME_EPOCH_SECONDS is $START_TIME_EPOCH_SECONDS"
+((END_TIME_EPOCH_SECONDS= "$START_TIME_EPOCH_SECONDS" + "$SOAK_TEST_DURATION_MINUTES"*60))
+echo "[INFO] END_TIME_EPOCH_SECONDS is $END_TIME_EPOCH_SECONDS"
+
+pushd "${CONTROLLER_E2E_PATH}" 1>/dev/null
+# Treat e2e directory as a module to run bootstrap as scripts
+export PYTHONPATH=..
+pip install -r requirements.txt
```
RedbackThomson, line 29: This is already being done in the Dockerfile?

soak/run_soak_test.sh
```diff
@@ -0,0 +1,44 @@
...
+#If current time is less than $END_TIME_EPOCH_SECONDS, keep executing the e2e tests.
+while [ $(date +%s) -le $END_TIME_EPOCH_SECONDS ]
+do
+  echo "[INFO] Current time is $(date +%s%3N) and  end time is $END_TIME_EPOCH_SECONDS in epoch seconds. Executing e2e tests..."
```
RedbackThomson, line 33: nit:
```suggestion
  echo "[INFO] Current time is $(date +%s%3N) and end time is $END_TIME_EPOCH_SECONDS in epoch seconds. Executing e2e tests..."
```

soak/run_soak_test.sh
```diff
@@ -0,0 +1,44 @@
...
+  python service_bootstrap.py
```
RedbackThomson, line 34: Bootstrapping and cleanup only needs to be done once. Move outside of while loop

RedbackThomson (changes requested):
A great start on the documentation. I have some concerns over the sheer number of manual steps. I understand many of these are limitations of the products, but I imagine a lot of these can be set in helm values or done with scripts that would save hassle. 

Also this PR seems to have pushed some `.DS_Store` files (Mac hidden files) - I'd recommend adding these to your global `.gitignore` file.

soak/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+# Persist build arguments into environment variables
+ARG AWS_SERVICE
+ARG E2E_GIT_REF=main
+ARG CONTROLLER_E2E_PATH=./${AWS_SERVICE}-controller/test/e2e
+ENV AWS_SERVICE ${AWS_SERVICE}
+ENV PYTHONPATH ${CONTROLLER_E2E_PATH}
+ENV CONTROLLER_E2E_PATH ${CONTROLLER_E2E_PATH}
+
+WORKDIR /
```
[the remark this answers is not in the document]
A-Hilaly: `WORKDIR` creates the target directory if doesn't exist https://docs.docker.com/engine/reference/builder/#workdir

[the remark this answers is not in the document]
A-Hilaly: I would also prefer using absolute paths and avoid using tildes, some docker build commands don't support tildes (`COPY` for example).
For this case i'm thinking we can use something like `WORKDIR /soak` or `WORKDIR /test-infra/soak` or maybe `/github.com/aws-controllers-k8s/test-infra/soak`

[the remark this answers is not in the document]
vijtrip2: Using /soak path instead of root directory. Thanks!

soak/README.md
vijtrip2: >"soak tests are run before we release into the public chart repo"

This will only be true for stable release. For patch releases (semver update), soak tests will run after the release and before marking the release as stable.

vijtrip2:
> A great start on the documentation. I have some concerns over the sheer number of manual steps. I understand many of these are limitations of the products, but I imagine a lot of these can be set in helm values or done with scripts that would save hassle.
> 
> Also this PR seems to have pushed some `.DS_Store` files (Mac hidden files) - I'd recommend adding these to your global `.gitignore` file.

Thank you! Very good suggestions!

Based on the feedback, I am working today to minimize number of copy-paste service team have to do.
Also separating the configuration variables(export commands) into a single place. (soak-test configuration file.)

soak/run_soak_test.sh
vijtrip2: This is a different requirements.txt

The one in Dockerfile is from test-infra repository. This one is from service-controller test directory.

RedbackThomson: You don't need to run both. The one in the service-controller pulls a specific version of the `test-infra` as part of its dependency tree. See SageMaker's for example: https://github.com/aws-controllers-k8s/sagemaker-controller/blob/main/test/e2e/requirements.txt

soak/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+ENV PYTHONPATH ${CONTROLLER_E2E_PATH}
+ENV CONTROLLER_E2E_PATH ${CONTROLLER_E2E_PATH}
+
+WORKDIR /soak
+# Install dependencies for soak test environment
+RUN apk add -y --no-cache git bash
+
+# Install python dependencies for running ack e2e tests
+COPY requirements.txt .
+RUN pip install -r requirements.txt
```
RedbackThomson, line 17: As I stated below in another comment, the `requirements.txt` dependencies in the `test-infra` repository are already included as part of the test dependency graph of the service-controller e2e tests. We actually don't want to install them here, because the service-controller e2e tests may be referencing an older version of the `test-infra` `acktest` package.

Might I suggest that after we clone (on L26), we `pip install` the requirements there. This way the container does not need to install requirements at runtime i.e in the `run_soak_test.sh` script.

soak/README.md
```diff
@@ -0,0 +1,274 @@
+## Introduction
+
+ACK team provides the soak testing framework where service teams can execute long running soak tests for testing the
+performance and stability of the service controllers. Running soak tests is also a requirement for cutting the stable
+release for the service controller. Find more details on the release process [here](https://github.com/aws-controllers-k8s/community/blob/main/docs/contents/releases.md).
+
+Under the hood, the soak test introduces load by repeatedly running the end-to-end tests for the service controller. End
+to end tests are present in service-controller github repository. i.e. "https://github.com/aws-controllers-k8s/<service-name>-controller".
+
+If you wish to run soak tests for your service controller locally, see the "HOW TO RUN SOAK TEST LOCALLY" section below.
```
RedbackThomson, line 10: nit:
```suggestion
If you wish to run soak tests for your service controller locally, see the "How To Run Soak Test Locally" section below.
```

soak/README.md
```diff
@@ -0,0 +1,274 @@
...
+```
+# NOTE: Substitute all the <placeholders> with actual values.
+# A handy alias to go to root of ACK soak test directory
+alias go-to-soak='cd ~/.ack/soak'
+
+# SERVICE CONTROLLER CONFIGURATION
+# AWS Service name of the controller under test. Ex: apigatewayv2
+export SERVICE_NAME=<aws-service-name>
+# Semver version of the controller under test. Ex: v0.0.2
+export CONTROLLER_VERSION=<semver>
```
RedbackThomson, line 51: This is the Github tag that has been released?

soak/README.md
```diff
@@ -0,0 +1,274 @@
...
+export CONTROLLER_CHART_URL=public.ecr.aws/aws-controllers-k8s/chart:$SERVICE_NAME-$CONTROLLER_VERSION
+# Release name of controller helm chart. Ex: soak-test
+export CONTROLLER_CHART_RELEASE_NAME=<controller-chart-release-name>
+
+# PROMETHEUS, GRAFANA CONFIGURATOIN
+# Release name of kube-prometheus helm chart. Ex: prom
+export PROM_CHART_RELEASE_NAME=<kube-prometheus-chart-release-name>
```
RedbackThomson, line 58: nit: Provide a sane default - I don't think any team cares too much to customise the chart release name.

soak/README.md
```diff
@@ -0,0 +1,274 @@
...
+# Local port to access Prometheus dashbaord
+export LOCAL_PROMETHEUS_PORT=9090
+# Local port to access Prometheus dashbaord
+export LOCAL_GRAFANA_PORT=3000
+
+# SOAK TEST RUNNER CONFIGURATION
+# Git branch of service controller repository where e2e tests are present.
+export E2E_GIT_REF=main
```
RedbackThomson, line 66: This should be the git commit of the tag (`CONTROLLER_VERSION`)? Otherwise tests may be out of sync with the release version

soak/README.md
```diff
@@ -0,0 +1,274 @@
...
+# Image repository where soak test runner image will be stored. (This can be your personal repository.) Ex: 'my-ack-repo/soak-test'
```
RedbackThomson, line 67: nit: This needs to be accessible by the EKS cluster - so should be a public ECR repo?

soak/README.md
```diff
@@ -0,0 +1,274 @@
...
+go-to-soak \
+&& helm repo add prometheus-community https://prometheus-community.github.io/helm-charts \
+&& helm install $PROM_CHART_RELEASE_NAME prometheus-community/kube-prometheus-stack \
+&& kubectl port-forward service/$PROM_CHART_RELEASE_NAME-kube-prometheus-stack-prometheus $LOCAL_PROMETHEUS_PORT:9090 & \
+&& kubectl port-forward service/$PROM_CHART_RELEASE_NAME-grafana $LOCAL_GRAFANA_PORT:80 &
+```
+
+NOTE:
+* Prometheus, Grafana services will be started as ClusterIP services and exposed on `localhost` using `kubectl port-forward`.
+* The port-forward processes will run in the background and steps to clean them up are mentioned later in this guilde.
```
RedbackThomson, line 146: nit: `s/guilde/guide/`

RedbackThomson (changes requested):

soak/README.md
vijtrip2: Yes.

soak/README.md
vijtrip2: Fair point.

soak/README.md
vijtrip2: Good point but not just ECR. Any public repository is fine. I have added a note in description.

soak/run_soak_test.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+PYTEST_LOG_LEVEL=$(echo "${PYTEST_LOG_LEVEL:-"info"}" | tr '[:upper:]' '[:lower:]')
```
A-Hilaly, line 5: Maybe it's better to default `PYTEST_LOG_LEVEL` to "debug" ?

soak/helm/ack-soak-test/templates/job.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: batch/v1
+kind: Job
```
A-Hilaly, line 2: I thought that the goal was keep these tests running for ever/periodically. Maybe we could use a `batch/v1.CronJob` instead?

A-Hilaly (commented):
Very nice progress @vijtrip2 ! LGTM Overall 👍 
I left two questions below

soak/run_soak_test.sh
vijtrip2: i think info is good enough. These will just be test runs for >24 hrs. Having those many debug logs can be nuisance.
With info level, I believe all the info and detailed error logs are printed whenever pytest failure is there.

Why did you think debug will be a better choice?

soak/helm/ack-soak-test/templates/job.yaml
vijtrip2: No, that's not the idea.
These tests will run once for a longer duration.

soak/run_soak_test.sh
A-Hilaly: > With info level, I believe all the info and detailed error logs are printed whenever pytest failure is there.

If that's the case i agree that `info` will be better!

> Why did you think debug will be a better choice?

I just thought that the logs will not show `debug` logs if a controller fails

PR #51, opened 2021-06-08
Optional SSM agent and PVRE reporting

RedbackThomson:
Installs an SSM agent installer `DaemonSet` into the cluster that ensures all nodes have an Amazon SSM agent - taken from https://docs.aws.amazon.com/prescriptive-guidance/latest/patterns/install-ssm-agent-on-amazon-eks-worker-nodes-by-using-kubernetes-daemonset.html

Creates an SSM association for all cluster instances and a respective SSM DataSync for instance inventory collection.

### DIFF
--- a/prow/ssm-agent/kustomization.yaml
+++ b/prow/ssm-agent/kustomization.yaml
@@ -2,7 +2,4 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
 resources:
-- prow-config.yaml
-- prow-data-plane.yaml
-- prow-jobs.yaml
-- ssm-agent.yaml
+- ssm-DaemonSet.yaml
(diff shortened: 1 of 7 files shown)


prow/ssm-agent/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+- ssm-DaemonSet.yaml
```
jaypipes, line 5: nit: lowercase `ssm-daemonset.yaml`

RedbackThomson: Kept it inline with the templates in the other Helm charts - https://github.com/aws-controllers-k8s/test-infra/tree/main/prow/data-plane/templates

jaypipes: meh, ok :)

PR #55, opened 2021-06-16
Increase prowjob timeout to 2 days for running soak tests.

vijtrip2:
* Increase prowjob timeout to 2 days from default 2 hours so that we can execute long running soak tests from prow cluster
* Use default namespace instead of 'test-pods' for installing helm charts in service team's soak cluster.

These were the two bugs found while I successfully executed soak test for ecr controller on a local prow cluster.

### DIFF
--- a/soak/helm/ack-soak-test/templates/job.yaml
+++ b/soak/helm/ack-soak-test/templates/job.yaml
@@ -2,6 +2,7 @@ apiVersion: batch/v1
 kind: Job
 metadata:
   name: {{ .Values.awsService }}-soak-test
+  namespace: {{ .Release.Namespace }}
   labels:
     app: {{ .Values.awsService }}-soak-test
 spec:
--- a/soak/prow/scripts/soak-on-release.sh
+++ b/soak/prow/scripts/soak-on-release.sh
@@ -81,6 +81,8 @@ export AWS_ACCOUNT_ID
 >&2 echo "soak-on-release.sh] [SETUP] Exported ACCOUNT_ID."
 
 aws eks update-kubeconfig --name soak-test-cluster >/dev/null
+# Use 'default' namespace by default and not 'test-pods' during helm commands.
+kubectl config set-context --current --namespace=default >/dev/null
 >&2 echo "soak-on-release.sh] [INFO] Updated the kubeconfig to communicate with 'soak-test-cluster' eks cluster."
 
 export HELM_EXPERIMENTAL_OCI=1
--- a/soak/prow/scripts/soak-on-release.sh
+++ b/soak/prow/scripts/soak-on-release.sh
@@ -138,8 +138,10 @@ helm install $SOAK_CHART_RELEASE_NAME . \
 # Loop until the Job executing soak test does not complete. Check again with 30 minutes interval.
 while kubectl get jobs/$AWS_SERVICE-soak-test -o=json | jq -r --exit-status '.status.completionTime'>/dev/null; [ $? -ne 0 ]
 do
-  >&2 echo "soak-on-release.sh] [INFO] Completion time is not present in the job status. Soak test is still running."
-  >&2 echo "soak-on-release.sh] [INFO] Sleeping for 30 mins..."
+  >&2 echo "soak-on-release.sh] [INFO] Current soak Job(default/$AWS_SERVICE-soak-test) status is: "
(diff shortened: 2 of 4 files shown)


soak/helm/ack-soak-test/templates/job.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: batch/v1
 kind: Job
 metadata:
   name: {{ .Values.awsService }}-soak-test
+  namespace: {{ .Release.Namespace }}
```
RedbackThomson, line 5: Hmm this is odd. This shouldn't be necessary for Helm 3 if setting `--namespace` in each of the commands.

RedbackThomson (commented):
I had a look at the log output from the currently running soak tests on the development cluster. Some things that I think would be nice for readability of the logs:

- Add timestamps for the waiting part (so we can be sure the pod didn't freeze)
- Print the soak pod status or even just a `kubectl get ...` for diagnosing wait times
- Print out the details of the soak pod (name, namespace) for debugging purposes

Also the `values.yaml` contains account IDs, so we probably can't print those. At least not without filtering them through yq.

vijtrip2:
> * Add timestamps for the waiting part (so we can be sure the pod didn't freeze)

Good idea, I will add this.

> *Print the soak pod status or even just a kubectl get ... for diagnosing wait times
> * Print out the details of the soak pod (name, namespace) for debugging purposes

You mean 'job'? We use job which ends up scheduling pods. And Yes, I will add it.

>"Also the values.yaml contains account IDs, so we probably can't print those. At least not without filtering them through yq."

The values.yaml print was removed when raising this PR. It was for my debugging purposes. That run had many debug statements as I was trying to make things work. They are not here in the PR.

soak/helm/ack-soak-test/templates/job.yaml
vijtrip2: We are not setting --namespace in each of the command. I tried doing that but it was being ignored. There is a bug in helm repo https://github.com/helm/helm/issues/3553

That's how I found about `kubectl config --namespace` command and it worked.

RedbackThomson: This is regarding `helm template`. I see we are using `helm install`, which should not have this bug?

soak/prow/scripts/soak-on-release.sh
```diff
@@ -138,8 +138,10 @@ helm install $SOAK_CHART_RELEASE_NAME . \
 # Loop until the Job executing soak test does not complete. Check again with 30 minutes interval.
 while kubectl get jobs/$AWS_SERVICE-soak-test -o=json | jq -r --exit-status '.status.completionTime'>/dev/null; [ $? -ne 0 ]
 do
-  >&2 echo "soak-on-release.sh] [INFO] Completion time is not present in the job status. Soak test is still running."
-  >&2 echo "soak-on-release.sh] [INFO] Sleeping for 30 mins..."
+  >&2 echo "soak-on-release.sh] [INFO] Current soak Job(default/$AWS_SERVICE-soak-test) status is: "
+  kubectl get jobs/$AWS_SERVICE-soak-test -o=json | jq -r '.status' >&2
+  >&2 echo "soak-on-release.sh] [INFO] Completion time is not present in the soak Job status. Soak Job is still running."
+  >&2 echo "soak-on-release.sh] [INFO] Current time is $(date) . Sleeping for 30 mins ..."
```
RedbackThomson, line 144: Why `echo` to `stderr`?

vijtrip2: Funny you ask. I learnt it from your code of [wrapper.sh](https://github.com/aws-controllers-k8s/test-infra/blob/main/prow/jobs/images/wrapper.sh) 

I assumed it is because if stdout is not printed in some systems to avoid noise but stderr is much more important and would be printed. Please correct me if there is a better reason for doing ">&2".

RedbackThomson: `wrapper.sh` was taken from upstream K8s and I don't particularly understand why they print everything to both pipes. For Prow the stdout shows in the logs anyway, so it doesn't make any difference.

vijtrip2: Okay, I will remove it.

PR #57, opened 2021-06-16
Parallelize e2e runs on the test-level

echen-98:
Currently each thread executes a single test file (`--dist loadfile`). In the case of the ElastiCache controller, tests are organized by resource. Since the Replication Group resource contains the most tests and the longest-running tests, both the actual tests and the teardown are prone to fail due to credential expiration during Prow runs. Using the option `--dist no` should mitigate this issue, which other controllers may eventually run into as well.

### DIFF
--- a/scripts/run-tests.sh
+++ b/scripts/run-tests.sh
@@ -84,7 +84,7 @@ elif [[ "$RUN_PYTEST_LOCALLY" == "true" ]]; then
     python service_bootstrap.py
     set +e
 
-    pytest -n $PYTEST_NUM_THREADS --dist loadfile -o log_cli=true \
+    pytest -n $PYTEST_NUM_THREADS --dist no -o log_cli=true \
       --log-cli-level "${PYTEST_LOG_LEVEL}" --log-level "${PYTEST_LOG_LEVEL}" .
     test_exit_code=$?
 


scripts/run-tests.sh
```diff
@@ -84,7 +84,7 @@ elif [[ "$RUN_PYTEST_LOCALLY" == "true" ]]; then
     python service_bootstrap.py
     set +e
 
-    pytest -n $PYTEST_NUM_THREADS --dist loadfile -o log_cli=true \
+    pytest -n $PYTEST_NUM_THREADS --dist no -o log_cli=true \
```
RedbackThomson, line 87: I'm struggling to understand from the [documentation](https://pypi.org/project/pytest-xdist/) what `no` implies. I would assume that would be no parallelism? 

The earliest tests I wrote relied on loading at a file level because the tests were expected to be completed in order. I see the SageMaker team has since updated them, but there might be other teams where this breaks. 

Either way, I think this is a good change for all of the tests to be able to run separately - without depending on the ordering within a file.

echen-98: I don't know what `no` is supposed to mean, but it's probably not the English word "no" since this is actually the option with the most parallelism.

If this will break other controllers' tests, then I think we would at least need some sort of support to configure this value for each service team during Prow runs. If we want parallel execution of tests then the `--dist loadfile` option seems too restrictive, since it limits us to one test case per file.

RedbackThomson: I agree `loadfile` is too restrictive. I'd be willing to apply this change and then if there are tests that need to be rewritten, we encourage the code owners of those to do so - for their own sake.

PR #61, opened 2021-06-17
Use soakconfig file for controlling soak test behavior

vijtrip2:
issue https://github.com/aws-controllers-k8s/community/issues/832
* Updating soak-runner image to include 'yq' tool and 'default_soak_config.yaml' file
* Adding a 'default_soak_config.yaml' file to provide default soak test behavior. (24 hours continuous e2e tests)
* Always pull soak-runner image in soak-cluster because this image is built and updated by prow on every soak run.
* Add instructions for service teams to configure custom soak behavior
* Execute pytest with marker option for providing custom behavior.

### DIFF
--- a/soak/default_soak_config.yaml
+++ b/soak/default_soak_config.yaml
@@ -1,0 +1,13 @@
+# Amount of time in minutes to execute long running soak test. Default 1 day = 1440 minutes
+durationMinutes: 1440
+# Map of pytest marker name to pytest command options
+# Soak test running container will execute pytest command for each of these markers sequentially from the e2e directory.
+# One example would be to run 'service' marker pytests followed by 'e2e_dangling_resource_cleanup'(example name) marker
+# to eliminate piling of dangling resources from e2e test runs, since soak test runner will be executing e2e tests
+# continuously for 24 hours.
+pytestMarkers:
+  # For default soak config, only run e2e tests continuously
+  service:
+    logLevel: info
+    numThreads: auto
+    dist: no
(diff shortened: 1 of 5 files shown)


soak/default_soak_config.yaml
```diff
@@ -0,0 +1,14 @@
+# Amount of time in minutes to execute long running soak test. Default 1 day = 1440 minutes
+durationMinutes: 1440
+# Map of pytest marker name to pytest command options
+# Soak test running container will execute pytest command for each of these markers sequentially from the e2e directory.
+# One example would be to run 'service' marker pytests followed by 'e2e_dangling_resource_cleanup'(example name) marker
+# to eliminate piling of dangling resources from e2e test runs, since soak test runner will be executing e2e tests
+# continuously for 24 hours.
+pytestMarkers:
+  # For default soak config, only run e2e tests continuously
+  service:
```
RedbackThomson, line 10: The default service marker for any given service is the name of the service itself - https://github.com/aws-controllers-k8s/rds-controller/blob/main/test/e2e/__init__.py#L25

I think providing `-m service` will be invalid

soak/default_soak_config.yaml
```diff
@@ -0,0 +1,14 @@
...
+    logLevel: info
+    numThreads: auto
+    dist: no
+    logCli: true
```
RedbackThomson, line 14: I think this should always be set to `true` - we should always be producing `stdout` for any soak tests.

soak/default_soak_config.yaml
vijtrip2: That is not the case, I have tested it. "-m service " works.

soak/default_soak_config.yaml
vijtrip2: I will update it.

jaypipes (approved):
Nice :) Thanks @vijtrip2!

PR #63, opened 2021-07-01
Migrate bootstrap resources to OOP style

RedbackThomson:
This pull request modifies how bootstrap resource creation and cleanup is handled, migrating from an imperative style to an object oriented style. Resources are now represented as an implementation of a base `BootstrappableResource` abstract type that declares only that it can be "boostrapped" and "cleaned up". Service teams should then override the `ServiceBootstrapResources` class, adding additional members for each bootstrappable resource. The `ServiceBootstrapResources` class handles serialization/deserialization (now into `pickle` instead of YAML) and calling the respective `bootstrap` and `cleanup` method for each member.

Services teams are encouraged to add additional `BootstrappleResource` types to the `test-infra` repository. Teams are also welcome to extend any `BootstrappleResource`, adding more fields or operations, in their own repositories if they require interactions specific to their service.

⚠️ This breaks backwards compatibility:
- `write_bootstrap_config` and `read_bootstrap_config` are now deprecated in favour of `BootstrappleResource` `serialize` and `deserialize` methods

Example of bootstrapping, now:
```python
resources = TestBootstrapResources(
    TestRole=Role("test-role", "eks.amazonaws.com", ["arn:aws:iam::aws:policy/AmazonEKSClusterPolicy"]),
    TestVPC=VPC(name_prefix="test-vpc", public_subnets=2, private_subnets=2)
)
resources.bootstrap()
resources.serialize(bootstrap_directory)
```

Example of cleanup, now:
```python
resources = ServiceBootstrapResources.deseralize(bootstrap_directory)
resources.cleanup()
```

### DIFF
--- a/src/acktest/bootstrapping/__init__.py
+++ b/src/acktest/bootstrapping/__init__.py
@@ -5,7 +5,7 @@ import pickle
 import logging
 
 from pathlib import Path
-from dataclasses import dataclass, fields
+from dataclasses import dataclass, fields, asdict
 from typing import Iterable, Iterator
 
 from ..aws.identity import get_region
@@ -46,28 +46,25 @@ class Serializable:
             bootstrap = pickle.load(stream)
         return bootstrap
 
+class BootstrapFailureException(Exception):
+    pass
 @dataclass
 class Bootstrappable(abc.ABC):
     """Represents a single bootstrappable resource.
     """
    
-    @abc.abstractmethod
-    def bootstrap(self):
-        pass
-
-    @abc.abstractmethod
-    def cleanup(self):
-        pass
-
     @property
     def region(self):
         return get_region()
 
-class BootstrapFailureException(Exception):
-    pass
+    @abc.abstractmethod
+    def bootstrap(self):
+        self._bootstrap_subresources()
 
-@dataclass
-class Resources(Serializable, Bootstrappable):
+    @abc.abstractmethod
+    def cleanup(self):
+        self._cleanup_subresources()
+
     @property
     def iter_bootstrappable(self) -> Iterator[Bootstrappable]:
         """Iterates over the values of each field that extends the 
@@ -77,61 +74,85 @@ class Resources(Serializable, Bootstrappable):
             Iterator[BootstrappableResource]: A field value.
         """
         for field in fields(self):
-            if not issubclass(field.type, Bootstrappable):
+            if not isinstance(field.type, type) or not issubclass(field.type, Bootstrappable):
                 continue
 
-            yield getattr(self, field.name)
+            attr = getattr(self, field.name)
+            if attr is None:
+                continue
 
-    def bootstrap(self):
-        """Runs the `bootstrap` method for every `BootstrappableResource` 
-            subclass in the bootstrap dictionary.
-        """
+            yield attr
 
+    def _bootstrap_subresources(self):
         bootstrapped = []
-        logging.info("🛠️ Bootstrapping resources ...")
         for resource in self.iter_bootstrappable:
-            exceptions = []
+            should_cleanup = True
+
+            resource_name = type(resource).__name__
+            logging.info(f"Attempting bootstrap {resource_name}")
             for _ in range(BOOTSTRAP_RETRIES):
                 try:
                     # Bootstrap and add to list of successes
                     resource.bootstrap()
+                    logging.info(f"Successfully bootstrapped {resource_name}")
                     bootstrapped.append(resource)
+                    should_cleanup = False
                     break
+                except BootstrapFailureException as ex:
+                    # Don't attempt to retry if we reached maximum retries beneath
+                    raise ex
                 except Exception as ex:
-                    exceptions.append(ex)
+                    logging.error(f"Exception while bootstrapping {resource_name}")
+                    logging.exception(ex)
+                    # Clean up any dependencies the first attempt made
+                    logging.info(f"Cleaning up dependencies created by {resource_name}")
+                    resource.cleanup()
+                    logging.info(f"Retrying bootstrapping {resource_name}")
                     continue
             else:
-                # Hit retry limit
-                logging.error(f"🚫 Exceeded maximum retries ({BOOTSTRAP_RETRIES}) for bootstrapping resource")
-                for ex in exceptions:
-                    logging.exception(ex)
+                logging.error(f"🚫 Exceeded maximum retries ({BOOTSTRAP_RETRIES}) for bootstrapping {resource_name}")
+            
+            if should_cleanup:
                 # Attempt to clean up successfully bootstrapped elements
                 self._cleanup_resources(bootstrapped)
-                raise BootstrapFailureException()
-                
+                raise BootstrapFailureException(f"Bootstrapping failed for resource type '{resource_name}'")
 
-    def cleanup(self):
-        """Runs the `cleanup` method for every `BootstrappableResource` 
-            subclass in the bootstrap dictionary.
-        """
-        logging.info("🧹 Cleaning up resources ...")
+    def _cleanup_subresources(self):
         self._cleanup_resources(self.iter_bootstrappable)
 
     def _cleanup_resources(self, resources: Iterable[Bootstrappable]):
         # Iterate through list in reverse order, so that resources created last
         # (with the most dependencies) are the first to be deleted
         for resource in reversed(list(resources)):
-            exceptions = []
+            resource_name = type(resource).__name__
             for _ in range(CLEANUP_RETRIES):
                 try:
                     # Clean up and add to list of successes
+                    logging.info(f"Attempting cleanup {resource_name}")
                     resource.cleanup()
+                    logging.info(f"Successfully cleaned up {resource_name}")
                     break
                 except Exception as ex:
-                    exceptions.append(ex)
+                    logging.error(f"Exception while cleaning up {resource_name}")
+                    logging.exception(ex)
                     continue
             else:
                 # Hit retry limit
-                logging.error(f"🚫 Exceeded maximum retries ({BOOTSTRAP_RETRIES}) for cleaning up resource")
-                for ex in exceptions:
-                    logging.exception(ex)
+                logging.error(f"🚫 Exceeded maximum retries ({BOOTSTRAP_RETRIES}) for cleaning up {resource_name}")
+                logging.error(f"Possibly dangling resource ({resource_name}): {asdict(resource)}")
+
+@dataclass
+class Resources(Serializable, Bootstrappable):
+    def bootstrap(self):
+        """Runs the `bootstrap` method for every `BootstrappableResource` 
+            subclass in the bootstrap dictionary.
+        """
+        logging.info("🛠️ Bootstrapping resources ...")
+        self._bootstrap_subresources()                
+
+    def cleanup(self):
+        """Runs the `cleanup` method for every `BootstrappableResource` 
+            subclass in the bootstrap dictionary.
+        """
+        logging.info("🧹 Cleaning up resources ...")
+        self._cleanup_subresources()
--- a/src/acktest/bootstrapping/__init__.py
+++ b/src/acktest/bootstrapping/__init__.py
@@ -6,9 +6,12 @@ import logging
 
 from pathlib import Path
 from dataclasses import dataclass, fields
-from typing import Iterator
+from typing import Iterable, Iterator
 
 from ..aws.identity import get_region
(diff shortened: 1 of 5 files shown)


src/acktest/bootstrapping/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+        with open(path, "wb") as stream:
+            pickle.dump(self, stream)
+        logging.info(f"Wrote bootstrap to {path}")
+
+    @classmethod
+    def deseralize(cls, config_dir: Path, bootstrap_file_name: str = "bootstrap.pkl") -> ServiceBootstrapResources:
+        """ Reads a service bootstrap from a given bootstrap pickle file.
+
+        Args:
+            config_dir: The directory in which the bootstray yaml exists.
```
mbaijal, line 34: typo: s/bootstray/bootstrap/g

src/acktest/bootstrapping/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+@dataclass
+class ServiceBootstrapResources(abc.ABC):
+    """Represents a list of all bootstrappable resources required for a given
+    service's tests.
+    """
+    def serialize(self, output_path: Path, bootstrap_file_name: str = "bootstrap.pkl"):
+        """ Dumps the service bootstrap into a pickle file at a given path.
+
+        Args:
+            bootstrap: The bootstrap object.
```
mbaijal, line 20: extra arg ?

src/acktest/bootstrapping/iam.py
```diff
@@ -0,0 +1,94 @@
...
+            )
+
+        iam_resource = iam.get_role(RoleName=role_name)
+        resource_arn = iam_resource["Role"]["Arn"]
+
+        # There appears to be a delay in role availability after role creation
+        # resulting in failure that role is not present. So adding a delay
+        # to allow for the role to become available
+        time.sleep(ROLE_CREATE_WAIT_IN_SECONDS)
+        logging.info(f"Created role {resource_arn}")
```
mbaijal, line 68: Is there a plan to add the serialization to these bootstrap methods as discussed for this issue - https://github.com/aws-controllers-k8s/community/issues/845 ?

src/acktest/bootstrapping/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+import abc
+import pickle
+import logging
+
+from pathlib import Path
+from dataclasses import dataclass, fields
+from typing import Iterator
+
+@dataclass
+class ServiceBootstrapResources(abc.ABC):
```
jaypipes, line 12: how about we call this just `Serializable`, make it a non-abstract base class with only the `serialize` and `deserialize` methods on it.

src/acktest/bootstrapping/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+            if not issubclass(field.type, BootstrappableResource):
+                continue
+
+            yield getattr(self, field.name)
+
+    def bootstrap(self):
+        """Runs the `bootstrap` method for every `BootstrappableResource` 
+            subclass in the bootstrap dictionary.
+        """
+        logging.info("🛠️ Bootstrapping resources ...")
```
jaypipes, line 64: hrmph. emojis.

src/acktest/bootstrapping/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+from pathlib import Path
+from dataclasses import dataclass, fields
+from typing import Iterator
+
+@dataclass
+class ServiceBootstrapResources(abc.ABC):
+    """Represents a list of all bootstrappable resources required for a given
+    service's tests.
+    """
+    def serialize(self, output_path: Path, bootstrap_file_name: str = "bootstrap.pkl"):
```
jaypipes, line 16: should be a `@classmethod` like `deserialize`, yes?

src/acktest/bootstrapping/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+        for resource in self.bootstrappable_field_values:
+            resource.bootstrap()
+
+    def cleanup(self):
+        """Runs the `cleanup` method for every `BootstrappableResource` 
+            subclass in the bootstrap dictionary.
+        """
+        logging.info("🧹 Cleaning up resources ...")
+        for resource in self.bootstrappable_field_values:
+            resource.cleanup()
```
jaypipes, line 74: And then I'd recommend moving these to a new class called just `Resources` that implements `Serializable` and `Bootstrappable`:

```python
class Resources(Serializable, Bootstrappable):
    @property
    def iter_resources(self) -> Iterator[Bootstrapple]:
        ...
    def bootstrap(self):
        ...
    def cleanup(self):
        ...
```

src/acktest/bootstrapping/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+            ServiceBootstrapResources: The servicebootstrap resources read from
+                the file.
+        """
+        path = config_dir / bootstrap_file_name
+        with open(path, "rb") as stream:
+            bootstrap = pickle.load(stream)
+        return bootstrap
+
+    @property
+    def bootstrappable_field_values(self) -> Iterator[BootstrappableResource]:
```
jaypipes, line 47: how about calling this `iter_resources` (see below comment...)

src/acktest/bootstrapping/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+    def cleanup(self):
+        """Runs the `cleanup` method for every `BootstrappableResource` 
+            subclass in the bootstrap dictionary.
+        """
+        logging.info("🧹 Cleaning up resources ...")
+        for resource in self.bootstrappable_field_values:
+            resource.cleanup()
+
+@dataclass
+class BootstrappableResource(abc.ABC):
```
jaypipes, line 77: recommend renaming to just `Bootstrappable`

src/acktest/bootstrapping/iam.py
```diff
@@ -0,0 +1,94 @@
...
+        role_name = re.match(ROLE_ARN_REGEX, self.arn).group(1)
+        managedPolicy = iam.list_attached_role_policies(RoleName=role_name)
+        for each in managedPolicy["AttachedPolicies"]:
+            iam.detach_role_policy(RoleName=role_name, PolicyArn=each["PolicyArn"])
+
+        inlinePolicy = iam.list_role_policies(RoleName=role_name)
+        for each in inlinePolicy["PolicyNames"]:
+            iam.delete_role_policy(RoleName=role_name, PolicyName=each)
+
+        instanceProfiles = iam.list_instance_profiles_for_role(RoleName=role_name)
```
jaypipes, line 87: please be consistent in variable names :) use `snake_case` in Python instead of `camelCase`

src/acktest/bootstrapping/s3.py
```diff
@@ -0,0 +1,48 @@
...
+    name_prefix: str
+
+    # Outputs
+    name: str = field(init=False)
+    
+    def bootstrap(self):
+        """Creates an S3 bucket with an auto-generated name.
+        """
+        region = get_region()
+        self.name = resources.random_suffix_name(self.name_prefix, 63)
```
jaypipes, line 26: What if the caller wants to specify a particular name and not have it be randomized?

src/acktest/bootstrapping/s3.py
```diff
@@ -0,0 +1,48 @@
...
+    # Inputs
+    name_prefix: str
+
+    # Outputs
+    name: str = field(init=False)
+    
+    def bootstrap(self):
+        """Creates an S3 bucket with an auto-generated name.
+        """
+        region = get_region()
```
jaypipes, line 25: might be useful in the future to have `region` be a field on the Bucket resource instead of always pulled from `get_region`. This way, multi-region resources can be created in the same bootstrapping set.

src/acktest/bootstrapping/vpc.py
```diff
@@ -0,0 +1,185 @@
...
+from . import BootstrappableResource
+from .. import resources
+from ..aws.identity import get_region
+
+# Subnets inside the default VPC CIDR block will be of form 10.0.*.0/24
+VPC_CIDR_BLOCK = "10.0.0.0/16"
+
+
+@dataclass
+class VPC(BootstrappableResource):
```
jaypipes, line 17: This sets up a lot more than just a VPC. It sets up subnets, internet gateways, route tables, etc. This isn't always required for some simple fixtures. Recommend renaming this resource `VPCPublicInternet` or similar and having a simpler `VPCPrivate` that doesn't set up the public stuff.

src/acktest/bootstrapping/vpc.py
```diff
@@ -0,0 +1,185 @@
...
+        for subnet in self.public_subnet_ids + self.private_subnet_ids:
+            ec2.delete_subnet(SubnetId=subnet)
+
+            logging.info(f"Deleted subnet {self.name}")
+
+        ec2.delete_route_table(RouteTableId=self.private_route_table_id)
+        ec2.delete_route_table(RouteTableId=self.public_route_table_id)
+
+        vpc.detach_internet_gateway(InternetGatewayId=self.internet_gateway_id)
+        ec2.delete_internet_gateway(InternetGatewayId=self.internet_gateway_id)
```
jaypipes, line 114: for these methods that create multiple resources, how are you handling errors and cleaning up resources created before an error occurs? For instance, if you successfully create the VPC, route table and public subnets but then hit some API quota or resource limit when creating private subnets, are you planning to clean up the previously created route tables, public subnets and VPC? Generally I'm not seeing any error handling at all for either `bootstrap` or `cleanup`. Is that on purpose?

jaypipes (changes requested):
Awesome start on this @RedbackThomson! :) Left some suggestions and questions inline for you!

src/acktest/bootstrapping/__init__.py
RedbackThomson: 100% did it on purpose

src/acktest/bootstrapping/__init__.py
RedbackThomson: No I wanted this to look like `resources.serialize(...)`. Whereas deserialize is like a factory for these types.

src/acktest/bootstrapping/s3.py
RedbackThomson: I highly don't recommend that. That means you can bootstrap for at least one test per account, limiting you to run only a single integration test at a time. Would there be a reason to have a completely deterministic resource name?

src/acktest/bootstrapping/__init__.py
jaypipes: sorry, I'm not following you... could you elaborate?

src/acktest/bootstrapping/s3.py
jaypipes: I don't disagree with you. I was asking whether it would be possible to give it a particular name instead of always be randomized.

src/acktest/bootstrapping/__init__.py
RedbackThomson: In my mind: `Serializable` should represent a class that can be serialized and therefore knows how to serialize itself - i.e. calling `resource.serialize()`.
Changing this to a `@classmethod` would work, but I think `Serializable.serialize(resource)` looks as clean.

src/acktest/bootstrapping/s3.py
RedbackThomson: Yes it would be possible, but I pretty explicitly gated it so that it wouldn't be. We could easily change `name_prefix` to `name` and make it part of the constructor like any other field.

src/acktest/bootstrapping/s3.py
RedbackThomson: Best I can do at the moment is to have a `@property` field that can be overridden. `dataclass` does some funky stuff when you add default value fields in the subclass and it throws an error. There are work arounds but they're ugly. See https://stackoverflow.com/questions/51575931/class-inheritance-in-python-3-7-dataclasses

src/acktest/bootstrapping/vpc.py
RedbackThomson: I'm rewriting VPC so it's composed of a number of smaller `Bootstrappable` types. I don't think how it's written at the moment can solve that issue.

src/acktest/bootstrapping/__init__.py
jaypipes: k, fair enough. :)

PR #64, opened 2021-07-06
Abstract test class for adopted resources

RedbackThomson:
This pull requests creates an abstract base class that serves as the basis for any adopted resource tests. By default it contains a `test_smoke` test that simply creates the adopted resource, ensures it is adopted successfully and that the target resource was also created.

To implement an adopted resource test in the service tests, create a new test class that implements `AbstractAdoptionTest` as the base class then:
-  Override the `RESOURCE_PLURAL` and `RESOURCE_VERSION` class members
- Implement the `get_resource_spec` (and optionally the `bootstrap_resource` and `cleanup_resource`) methods


src/acktest/adoption.py
```diff
@@ -0,0 +1,144 @@
...
+ADOPTED_CONDITION_NAME = "ACK.Adopted"
+
+@dataclass(frozen=True)
+class AdoptedResourceAWSIdentifier:
+    """ Represents the base AWS identifier spec fields from the adopted resource CRD.
+
+    Additional keys need to be configured as allowlisted as part of the 
+    controller generator.
+    """
+    additionalKeys: Dict[str, str]
```
jaypipes, line 35: here and below, please ensure `snake_case` is used for field and variable names, per Python convention. You've used `snake_case` for most things here... just a few missed ones like here and `nameOrID` below.

src/acktest/adoption.py
```diff
@@ -0,0 +1,144 @@
...
+ADOPTED_RESOURCE_GROUP = "services.k8s.aws"
+ADOPTED_RESOURCE_VERSION = "v1alpha1"
+ADOPTED_RESOURCE_PLURAL = "adoptedresources"
+ADOPTED_RESOURCE_KIND = "AdoptedResource"
+
+ADOPTED_CONDITION_NAME = "ACK.Adopted"
+
+@dataclass(frozen=True)
+class AdoptedResourceAWSIdentifier:
+    """ Represents the base AWS identifier spec fields from the adopted resource CRD.
```
jaypipes, line 30: femto-nit: there an odd extra space character before the first word of the docstring... here and a number of other places below

src/acktest/adoption.py
```diff
@@ -0,0 +1,144 @@
...
+
+        return (reference, k8s.create_custom_resource(reference, body_dict))
+        
+
+    def _assert_adoption_status(self):
+        cr = k8s.wait_resource_consumed_by_controller(self._reference)
+
+        assert cr is not None
+        assert k8s.get_resource_exists(self._reference)
+        assert k8s.wait_on_condition(self._reference, ADOPTED_CONDITION_NAME, "True")
```
jaypipes, line 108: consider using the newly-added [`condition.assert_type_status`](https://github.com/aws-controllers-k8s/test-infra/blob/75752b2d9cd67aad898bf281c90f1d8b774391a5/src/acktest/k8s/condition.py#L28-L65) here:

```python
from .k8s import condition
...
condition.assert_type_status(self._reference, condition.CONDITION_TYPE_ADOPTED, True)
```

jaypipes: If you really want to wait with a timeout, I can add new waiter functions to that module...

src/acktest/adoption.py
```diff
@@ -0,0 +1,144 @@
...
+    kind: str
+
+@dataclass(frozen=True)
+class AdoptedResourceSpec:
+    """ Represents the adopted resource CRD spec fields.
+    """
+    aws: AdoptedResourceAWSIdentifier
+    kubernetes: AdoptedResourceKubernetesIdentifiers
+
+class AbstractAdoptionTest(ABC):
```
jaypipes, line 62: In the docstring for abstract base classes, I like to see example usage of the abc. Tell the readers how you expect your class to be instantiated and called in normal operation.

jaypipes (approved):
++ few tiny nits you can address in this PR or a followup. Nice work and sorry for the delay in reviewing, @RedbackThomson!

src/acktest/adoption.py
RedbackThomson: Can't do this. These values are used as the keys when serializing to and from JSON for k8s specs. We need them to have the correct case to match the CRD.

RedbackThomson:
@vijtrip2 and @jaypipes Please take a look. See https://github.com/aws-controllers-k8s/s3-controller/pull/65/commits/bb41ddaad1be974478b809620fef8d48bd180406 for an example of the implementation.

vijtrip2 (approved):
Looks very good. I have just one question.

src/acktest/k8s/resource.py
```diff
@@ -327,8 +365,7 @@ def wait_on_condition(reference: CustomResourceReference,
     if not desired_condition:
         logging.error(f"Resource {reference} does not have a condition of type {condition_name}.")
     else:
-        logging.error(f"Wait for condition {condition_name} to reach status {desired_condition_status} timed out")
-    logging.info(f"Resource state:\n{get_resource(reference)}")  # log resource state upon failure
```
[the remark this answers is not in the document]
jaypipes: it wasn't all that useful, IMHO. certainly did not represent "state upon failure" as the comment indicates...

jaypipes (approved):
:+1:

PR #65, opened 2021-07-08
Add ability to define wait_period and period_length in create call

ryansteakley:
Adds ability to specify `wait_period `and `period_length `in create call which will be passed to `wait_resource_consumed_by_controller`()

### DIFF
--- a/src/acktest/k8s/resource.py
+++ b/src/acktest/k8s/resource.py
@@ -78,7 +78,9 @@ def create_reference(crd_group: str,
     return reference
 
 def create_resource(reference: CustomResourceReference,
-                    spec: object):
+                    spec: object, 
+                    wait_periods: int, 
+                    period_length: int):
     """
     Create a resource from the reference and wait to be consumed by controller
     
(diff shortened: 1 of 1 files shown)


surajkota (approved):
/approve

RedbackThomson:
I would like to think the default wait period of 30 seconds is more than enough time for any resource to be consumed. I'd be a proponent for lowering it, but increasing it when tests fails seems to indicate there are bigger issues at play. Sleeping for longer is not a solution to waiting for poorly running code. Especially when running a small number of resources in e2e tests on a fresh cluster, there should be no resource that takes >30 seconds to be consumed once created.

surajkota:
This PR is to make the option configurable. Ideally yes we want to decrease it but right now we want to try different values to see observe the behaviour so we can arrive at a conclusion

jaypipes (approved):
Agree with @RedbackThomson that making wait periods longer than 30 seconds is probably an indication of bigger issues in the underlying tests/controller. However, that said, I see nothing wrong with making this stuff configurable.

PR #71, opened 2021-07-13
Publish controller logs as Prow artifact

RedbackThomson:
This pull request adds a new environment variable `$DUMP_CONTROLLER_LOGS` to the `kind-build-test.sh` script that optionally dumps the controller pod logs into the `$ARTIFACTS/controller_logs` folder. This folder is consumed by Prow and uploaded to S3 at the conclusion of the job.

It also configures `spyglass` with to add the `controller_logs` file to the `buildlog` lense, so that it is accessible from the web dashboard.


prow/config/templates/config-ConfigMap.yaml
```diff
@@ -52,15 +52,7 @@ data:
...
-          - build-log.txt
-        - lens:
-            name: junit
-          required_files:
-          - .*/junit.*\.xml
-        - lens:
-            name: podinfo
-          required_files:
-          - podinfo.json
+          - build-log.txt|controller_logs
```
jaypipes, line 55: is this right? You pipe directories to be created to the `required_files` value?

RedbackThomson: The documentation is ... nonexistant. From my tests this did work. I'm actually going to move it into the `optional_file` array, though. Since we may not reach that part of the script if tests fail early.

jaypipes (approved):
Nice :)

PR #72, opened 2021-07-14
Increase Prow node root disk size

RedbackThomson:
Modifies the managed nodegroup disk size from the 20GB default to 150GB. Attempting to fix the `no space left on device` error seen in some Prow jobs.


surajkota (approved):
/approve
/lgtm

jaypipes (approved):
Please note this is not EBS. This is the EC2 instance's root disk size.

PR #77, opened 2021-07-26
Use IAM role from AWS account hosting controller images for release-controller.sh script.

vijtrip2:
* Use IAM role from AWS account hosting controller images for release-controller.sh script.
* With new ecr repo migration, we need to use IAM role identity from the new account for publishing controller and chart images.

### DIFF
--- a/cd/scripts/release-controller.sh
+++ b/cd/scripts/release-controller.sh
@@ -76,6 +76,19 @@ fi
 
 echo "VERSION is $VERSION"
 
+ASSUME_EXIT_VALUE=0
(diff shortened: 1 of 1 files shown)


jaypipes (approved):
Nice, looks kosher to me.

jaypipes:
@vijtrip2 have you added the appropriate SSM parameter value? can I lgtm this?

vijtrip2:
> @vijtrip2 have you added the appropriate SSM parameter value? can I lgtm this?

Yes and Yes : )

PR #81, opened 2021-07-28
Publish CI test artifacts on non-zero exit

RedbackThomson:
Ensures the controller log artifact is pushed regardless of the exit code from `run-tests.sh`. This should allow service teams to debug their controller even in the event of a test failure.

### DIFF
--- a/scripts/kind-build-test.sh
+++ b/scripts/kind-build-test.sh
@@ -264,7 +264,9 @@ fi
 # run e2e tests
 export SKIP_PYTHON_TESTS
 export RUN_PYTEST_LOCALLY
+set +e
 $SCRIPTS_DIR/run-tests.sh $AWS_SERVICE
+EXIT_CODE=$?
 
 if [[ "$DUMP_CONTROLLER_LOGS" == true ]]; then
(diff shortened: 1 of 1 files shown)


vijtrip2 (approved):

scripts/kind-build-test.sh
```diff
@@ -264,7 +264,9 @@ fi
 # run e2e tests
 export SKIP_PYTHON_TESTS
 export RUN_PYTEST_LOCALLY
+set +e
 $SCRIPTS_DIR/run-tests.sh $AWS_SERVICE
+EXIT_CODE=$?
```
surajkota, line 270: -e after this, so if something else fails it gets caught

RedbackThomson: The only thing after this is dumping the controller logs. If that fails, we still want to continue execution until we reach the `exit $EXIT_CODE` line.

surajkota: sounds ok for now, if the actual tests passed and dumping logs fail, let the build pass

surajkota (approved):

PR #89, opened 2021-08-17
Support replacing custom resources

RedbackThomson:
This method allows resources to have their spec fields entirely replaced, rather than simply patched.


vijtrip2:
Your change looks good, but why does this PR have (my) commits that are already merged into the main?

PR #114, opened 2021-09-14
include ACK runtime version in auto-generated controller branch

vijtrip2:
Description of changes:
a) Choosing the dependabot approach for generating GitHub branch name with auto-generated controllers.
b) Having the same branch name caused this [issue](https://github.com/aws-controllers-k8s/ec2-controller/pull/9#discussion_r708492058)

There were two approaches to solve problem mentioned in (b)
1. Look for any open PR for branch "ack-bot/runtime" and update that PR title and body with latest runtime version. There will be only one PR open at anytime for ACK runtime library.
2. Approach used by [Dependabot](https://github.com/dependabot/dependabot-core/branches/all). Create separate PR and branch for package-version. PROs are that separate PR and changes for each runtime version. Only Con is cleanup of old branches which can done in future with a periodic job.

Hence I chose the approach which is consistent with dependabot for upgrading ACK runtime dependency.

### DIFF
--- a/cd/auto-generate/auto-generate-controllers.sh
+++ b/cd/auto-generate/auto-generate-controllers.sh
@@ -34,9 +34,6 @@ TEST_INFRA_DIR=$CD_DIR/..
 WORKSPACE_DIR=$TEST_INFRA_DIR/..
 CODEGEN_DIR=$WORKSPACE_DIR/code-generator
 
-DEFAULT_PR_SOURCE_BRANCH="ack-bot/runtime"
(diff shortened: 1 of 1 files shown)


RedbackThomson:
Just to clarify, this approach will now push a new branch for every runtime version? Will it still override the existing PR?

vijtrip2:
> Just to clarify, this approach will now push a new branch for every runtime version? Will it still override the existing PR?

Yes, for that runtime version, it will override existing PR without opening a new one. 
There will only be one PR per runtime version.

A-Hilaly (approved):

PR #119, opened 2021-09-16
Fix Prometheus rules and allowlist

RedbackThomson:
Description of changes:
The original document where I found the Prow dashboards and rules was for a much older version of `kube-state-metrics`. Some of the metrics it references in the rules have since been replaced by ones with different names. I have updated the names of these metrics, and fixed the syntax in the metrics allow list.


vijtrip2 (approved):
/lgtm

PR #121, opened 2021-09-17
New `code-generator` presubmit validations

vijtrip2:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/892

Description of changes:
* New presubmit prowjob on `code-generator` repo, which will regenerate `dynamodb-controller` locally and then perform `unit tests`, `e2e tests` and `helm tests` on the newly generated controller.
* Once this prowjob is successful and stable, I will add more service controllers as candidates for testing `code-generator` changes automatically before merging them
* Since sometimes it is necessary to make breaking changes in code-generator, these new prow-jobs/tests are optional for tide-merge. If they fail, `code-generator` maintainers can look at prowlogs before commenting 'lgtm'


prow/jobs/jinja/presubmits/code_gen_tests.jinja2
```diff
@@ -0,0 +1,58 @@
...
+      serviceAccountName: pre-submit-service-account
+      containers:
+      - image: public.ecr.aws/m5q3e4b2/prow:prow-integration-0.0.4
+        securityContext:
+          privileged: true
+        resources:
+          requests:
+            cpu: 2
+          limits:
+            cpu: 8
```
RedbackThomson, line 46: Can definitely remove this. These requests and limits are arbitrary at best, misleading at worst. I will probably remove them from the other e2e tests.

prow/jobs/jinja/presubmits/code_gen_tests.jinja2
```diff
@@ -0,0 +1,58 @@
+  aws-controllers-k8s/code-generator:
```
RedbackThomson, line 1: nit: Rename this file to `code_generator_tests.jinja2` for clarity

prow/jobs/jinja/presubmits/code_gen_tests.jinja2
```diff
@@ -0,0 +1,58 @@
...
+      serviceAccountName: pre-submit-service-account
+      containers:
+      - image: public.ecr.aws/m5q3e4b2/prow:prow-unit-0.0.1
+        resources:
+          limits:
+            cpu: 1
+        command: ["make", "test"]
+
+{% for service in code_gen_presubmit_services  %}
+  - name: {{ service }}-controller-test
```
RedbackThomson, line 16: One test for every single service is going to greatly increase the number of checks on every PR, and greatly increase the load on Prow. I'm fine with this for now but we may want to rethink how this scales, or how we can parallelise these loads, in the future.

RedbackThomson (changes requested):
I'm interested to see how this works, and the impact of these tests on our testing cluster over time. I would suggest that we disable e2e tests for this, at the moment. I think they are redundant, considering the PRs that are produced when a new code-generator version are released will trigger these tests, anyway. 

> Since sometimes it is necessary to make breaking changes in code-generator, these new prow-jobs/tests are optional for tide-merge. If they fail, code-generator maintainers can look at prowlogs before commenting 'lgtm'

++ For this. This is a "nice-to-have" but ultimately not a deal breaker for any of our changes.

prow/jobs/jinja/presubmits/code_gen_tests.jinja2
vijtrip2: Definitely. It does not make sense to run for all AWS services. But we will pick our selected controllers in future.

vijtrip2:
>  I think they are redundant, considering the PRs that are produced when a new code-generator version are released will trigger these tests, anyway.

I do not completely agree with this. The basis for adding these validations is to catch the failure before core library changes are merged. 
The e2e tests which you mentioned run as part of post-submit job.

Is it okay if we run these e2e tests and helm tests for now?
If we find no value in them or find that they are overwhelming our prow cluster, then we can get rid of them or run just a selected few?

RedbackThomson:
> I do not completely agree with this. The basis for adding these validations is to catch the failure before core library changes are merged.
The e2e tests which you mentioned run as part of post-submit job.

Please correct me if I am wrong, but are we not running the same tests as presubmit jobs in each of the service controllers? Here we build and test each controller with the new code-gen, with the service controllers we build and test it with the newest tag of code-gen (when one is pushed). 

I think there is still valid reason to run them now - that is early detection of critical errors (eg. every single service controller breaks). However it still feels redundant.

RedbackThomson (approved):
/lgtm

vijtrip2:
> Please correct me if I am wrong, but are we not running the same tests as presubmit jobs in each of the service controllers?

You are correct. It's same e2e tests.
Two reasons to run them again as part of service-controller PR are:
a) We will not include all service controllers here in code-generator presubmit
b) Failures detected here can be fixed in same PR(before code-gen release). If not, the fixes will require cutting another release of code-gen with bug fixes.

PR #137, opened 2021-10-06
Use lightweight aws-cli image for test-recommended-policy

vijtrip2:
Description of changes:
* reproduce the `/.aws permission denied` issue locally by running aws-cli image in a pod
* aws-cli container by default runs as non root container, and to run the aws command it tries to locate `/.aws` directory and gets permission denied
* This works fine in docker because default runAsUser for docker is root
*  This change runs aws-cli image as root for recommended-policy-arn test. 
* Our other images which use buster:slim as base also run as root by default
* Tested locally that running as root fixes the permission issue


RedbackThomson:
Thank you, this is preferable over running the heavier integration test containers. It is a shame that they need privileged permissions, but I understand this is out of our control.

/lgtm

prow/jobs/jinja/presubmits/service_tests.jinja2
```diff
@@ -88,7 +88,9 @@
     spec:
       serviceAccountName: pre-submit-service-account
       containers:
-      - image: public.ecr.aws/m5q3e4b2/prow:prow-integration-0.0.5
```
jaypipes, line 91: @vijtrip2 @RedbackThomson so does this mean we can delete the prow-integration image from our internal ack-infra ECR Public repos and this image definition file? https://github.com/aws-controllers-k8s/test-infra/blob/main/prow/jobs/images/Dockerfile.integration

PR #145, opened 2021-10-18
prevent race cond by moving crd deletion to before helm uninstall

bwagner5:
Issue #, if available:
N/A

Description of changes:
 - Prevent a race condition where CRDs are deleted after the helm chart is uninstalled which prevents the CRDs from being deleted since the controller is erroring out. This race condition caused the `kubectl delete crds...` command to freeze forever. To fix, I just moved the crd deletion to before the helm chart uninstall. 
 - Add `--wait` to helm install command to not rely as much on the 10sec wait. I think the controllers should have real readiness checks at some point to completely mitigate the need for any 10 sec wait.

**Update: More info on race condition**

The race condition occurs when a test errors and resources are not deleted. The controller is not uninstalled when the test fails and there are resources left. Maybe this is just a controller test problem and it should try harder to clean up everything even when a test fails. But uninstalling the helm chart first causes any resource left behind to be orphaned because the finalizers can't complete. This PR moves the CRD deletion before the controller deletion which cleans up any resources left by tests before it brings down the controller.

To Reproduce:

1. Change a test to fail in a controller e2e test suite (ec2-controller preferably)
1. `make kind-test`
1. Re-run with the same cluster `TMP_DIR.... make kind-test` You should see CRD delete hang because instances of the CRD exist and can't complete their finalizer.

### DIFF
--- a/scripts/test-helm.sh
+++ b/scripts/test-helm.sh
@@ -98,6 +98,7 @@ pushd "$HELM_DIR" 1> /dev/null
     --set image.repository="$IMAGE_REPO" \
     --set image.tag="$IMAGE_TAG" \
     --set metrics.service.create=true \
+    --wait \
     "$HELM_CHART_NAME" . 1>/dev/null || exit 1
   echo "ok."
 popd 1> /dev/null
@@ -152,11 +153,11 @@ then
   exit 1
 fi
 echo "ok."
-echo -n "test-helm.sh] uninstalling the Helm Chart $HELM_CHART_NAME in namespace $K8S_NAMESPACE ... "
-helm uninstall --namespace "$K8S_NAMESPACE" "$HELM_CHART_NAME" 1>/dev/null || exit 1
-echo "ok."
 echo -n "test-helm.sh] removing $AWS_SERVICE crds installed by Helm ... "
 kubectl delete -f "$HELM_DIR/crds" 1>/dev/null || exit 1
+echo "ok."
+echo -n "test-helm.sh] uninstalling the Helm Chart $HELM_CHART_NAME in namespace $K8S_NAMESPACE ... "
+helm uninstall --namespace "$K8S_NAMESPACE" "$HELM_CHART_NAME" 1>/dev/null || exit 1
 echo "ok."
 echo -n "test-helm.sh] deleting $K8S_NAMESPACE namespace ... "
 kubectl delete namespace "$K8S_NAMESPACE" 1>/dev/null || exit 1


scripts/test-helm.sh
```diff
@@ -98,6 +98,7 @@ pushd "$HELM_DIR" 1> /dev/null
     --set image.repository="$IMAGE_REPO" \
     --set image.tag="$IMAGE_TAG" \
     --set metrics.service.create=true \
+    --wait \
```
vijtrip2, line 101: this is not needed. The pod will never be in ready state because it fails to start due to lack of AWS creds.

AWS creds are added after the helm install and controller starts successfully after that.

scripts/test-helm.sh
```diff
@@ -152,12 +153,12 @@ then
   exit 1
 fi
 echo "ok."
-echo -n "test-helm.sh] uninstalling the Helm Chart $HELM_CHART_NAME in namespace $K8S_NAMESPACE ... "
-helm uninstall --namespace "$K8S_NAMESPACE" "$HELM_CHART_NAME" 1>/dev/null || exit 1
-echo "ok."
 echo -n "test-helm.sh] removing $AWS_SERVICE crds installed by Helm ... "
 kubectl delete -f "$HELM_DIR/crds" 1>/dev/null || exit 1
 echo "ok."
+echo -n "test-helm.sh] uninstalling the Helm Chart $HELM_CHART_NAME in namespace $K8S_NAMESPACE ... "
```
vijtrip2, line 159: Can you please explain what is this race condition? I have not seen this test ever fall into any race condition.

In the original code, once helm uninstall finishes, the controller pod is no longer there. And I am having hard time grasping why any controller error will prevent CRD deletion when there is no controller remaining.

bwagner5: The race condition occurs when a test errors and resources are not deleted. The controller is not uninstalled when the test fails and there are resources left. Maybe this is just a controller test problem and it should try harder to clean up everything even when a test fails. But uninstalling the helm chart first causes any resource left behind to be orphaned because the finalizers can't complete. This PR moves the CRD deletion before the controller deletion which cleans up any resources left by tests before it brings down the controller. 

To Reproduce:

1) Change a test to fail in a controller e2e test suite (ec2-controller preferably)
2) `make kind-test`
3) Re-run with the same cluster `TMP_DIR.... make kind-test` You should see CRD delete hang because instances of the CRD exist and can't complete their finalizer.

scripts/test-helm.sh
bwagner5: The tests do pass locally with this. 

I'm not sure I understand what you mean.  The `wait` expects the controller pod to get to a `Ready` state.  If credentials are not valid, then it will eventually fail. But in the normal case where credentials are valid, this is just better than only relying on the 10 sec sleep that is directly below this. 

I might be missing something here :)

scripts/test-helm.sh
vijtrip2: I thought about this and I do not think this will really solve the deleting crd problem. The problem will still persist because helm install will not cleanup any custom resources causing the crd deletion to hang.

>"Why are we deleting the crds in first place"

This was added only for local testing. Since we want to e2e-test the controller with the crds inside config/ directory, I delete the crds installed by helm test and then crds from config/crd get installed before e2e tests run.
This all does not matter for Prowjobs because these tests are executed in isolation with each other on prow in different kind clusters

Interestingly, if we do not delete crds in helm tests they will be updated when config/crd is installed using kustomize.

I will raise the PR for that and it will avoid this specific error when tests fail and developer reuses the KinD cluster to execute both helm and e2e tests together.(So specific)

PR #153, opened 2021-10-21
do not use SIGKILL for cleaning up rotate-aws-creds bg task

vijtrip2:
Description of changes:
* reproduced the issue locally and found out using SIGKILL does not give opportunity for subprocesses to be cleaned up. They endup getting owned by PID 1
* using default SIGTERM cleans up all the processes and sleep


vijtrip2:
merging to validate fix on prow. Please leave comment if any and I will address them.

PR #174, opened 2022-01-09
Add support for MapPublicIP on bootstrapped VPCs

RedbackThomson:
Description of changes:
The EKS controller tests need their public subnets to automatically map public IPv4 addresses to all new instances within the subnet. This property is not enabled by default, but instead requires a call to the `modify_subnet_attribute` operation, instead. I have added a new property, with a default of `True`, for this option.

### DIFF
--- a/src/acktest/bootstrapping/vpc.py
+++ b/src/acktest/bootstrapping/vpc.py
@@ -94,6 +94,7 @@ class Subnets(Bootstrappable):
     cidr_blocks: List[str]
     is_public: bool = True
     num_subnets: int = 1
+    map_public_ip: bool = True
 
     # Subresources
     route_table: RouteTable = field(init=False, default=None)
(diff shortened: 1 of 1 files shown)


src/acktest/bootstrapping/vpc.py
```diff
@@ -124,6 +125,11 @@ def bootstrap(self):
             subnet = vpc.create_subnet(CidrBlock=self.cidr_blocks[i], AvailabilityZone=region_azs[i % len(region_azs)])
             self.subnet_ids.append(subnet.id)
 
+            # Make a separate call to enable MapPublicIpOnLaunch since boto3
+            # does not accept it in the `create_subnet` parameter list
```
jaypipes, line 129: heh.

jaypipes (approved):

PR #179, opened 2022-01-31
aws-sdk-go upgrade automation proposal

vijtrip2:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1130

Description of changes:
* Small proposal describing the approach to automate the `aws-sdk-go` upgrades in ACK repos


docs/design/proposal/aws-sdk-upgrades.md
```diff
@@ -0,0 +1,142 @@
...
+### Release
+The actual process of upgrading the `go.mod` files with latest `aws-sdk-go` version
+and raising PRs.
+
+## Implementation Details
+
+### Trigger
+For the trigger, the periodic job option should be preferred that will
+run every day. 24 hours frequency is good enough to not delay verifying
+the new `aws-sdk-go` release.
```
RedbackThomson, line 71: Is this too frequently? I see from the [aws-sdk-go releases](https://github.com/aws/aws-sdk-go/releases) that we would be getting a new version most days. This would mean a new PR almost every day. Most new releases probably don't contain relevant changes to runtime or code-generator directly, and therefore we probably don't care to update runtime that often.

Is it possible to know whether a new version of the SDK impacts CARM or IAM?

RedbackThomson (changes requested):
I am totally down with the automating this process. I just want to make sure we are automating it to upgrade for useful reasons ie. not just because we "should". I have left an inline comment regarding update frequency.

docs/design/proposal/aws-sdk-upgrades.md
vijtrip2: Excellent point!
I do agree that we should not update runtime on every patch release of `aws-sdk-go` (which seems almost everyday. I missed this).

Probably updating on every minor releases will be a good start (for runtime -> code-gen -> controllers) because I can see that for core SDK changes like session, endpoint etc, `aws-sdk-go` team cut a new minor release.

I will think more on handling specific patch release(Service Client Updates) for specific controllers.

RedbackThomson: I'm hesitant to update the service controller's automatically. Or maybe teams could opt-in for that function. The reason is that a lot of service controller's will require updates to their hooks to support new spec fields properly. Probably most controller's shouldn't, for most of their resources, but some controllers (*cough S3 cough*) are notorious for specific fields having "custom logic". I'd love to get service team's feedback on how useful this would be for them.

vijtrip2: > I'm hesitant to update the service controller's automatically

Are you suggesting to only update core libraries?
If we update the ACK `runtime` with some `aws-sdk-go` version, service controllers will need to be updated with at least that `aws-sdk-go` version when they upgrade to latest runtime.

> a lot of service controller's will require updates to their hooks to support new spec fields properly

This is exact reason for adding verification step, which provides insight into the concerns you raised. And the automation just raises the PR, if existing functionality gets broken, i.e. e2e tests fail, the PR does not get merged. If any new resources get added, the PR does not get merged.

docs/design/proposal/dependency-upgrades/aws-sdk-upgrades.md
```diff
@@ -0,0 +1,152 @@
...
+* aws-sdk-go **v1** has two kinds of releases. **Minor version updates** which
+mainly reflect the changes in **SDK Features**(enabling FIPS endpoints) and
+**Patch version updates** which mainly reflect **Service Client Updates**(updates
+in specific AWS service API + documentation). See this
+[v1.42.0 changelog](https://github.com/aws/aws-sdk-go/releases/tag/v1.42.0) as an example
+
+## Scope
+This documentation will focus on automating the updates for aws-sdk-go minor version
+upgrades. Minor version upgrades will be done starting from ACK runtime to code-generator
+and then to all service-controllers using existing automation.
```
RedbackThomson, line 47: ++ Nice scope change. Thank you.

docs/design/proposal/dependency-upgrades/aws-sdk-upgrades.md
```diff
@@ -0,0 +1,152 @@
...
+### Release
+The actual process of upgrading the go.mod files with latest aws-sdk-go version
+and raising PR.
+
+## Implementation
+
+### Trigger
+For the trigger, the periodic job option should be preferred that will
+run every day. 24 hours frequency is good enough to not delay verifying
+the new aws-sdk-go release.
```
RedbackThomson, line 78: I would add a filter here to ignore any releases that are only patches - passing only minor version updates through to verification. There is no point going any further with testing past this point otherwise. This should be as simple as checking if the patch version of the SDK semver is 0. Whenever you update the minor version, you have to reset the patch to 0.

docs/design/proposal/dependency-upgrades/aws-sdk-upgrades.md
```diff
@@ -0,0 +1,152 @@
...
+The testing will include updating the dependency, generating new controller
+source code, compiling and publishing a report of the upgrade. 
+The report will include details like,
+* Which controllers were successfully built and unit tested
+* Which controllers failed to build and compile 
+* The summary of files added/changed for each successful controller. This
+will give an idea where new resources/fields were added in new aws-sdk-go
+release
+
+###### Setup
```
RedbackThomson, line 95: This is a few too many indentations.
```suggestion
#### Setup
```

docs/design/proposal/dependency-upgrades/aws-sdk-upgrades.md
```diff
@@ -0,0 +1,152 @@
...
+This job will checkout all the service controller repos, along with latest
+ACK runtime release and corresponding code-generator release.
+
+> NOTE: We will not checkout `main` branch of ACK core libraries because this
+job should only validate the changes caused by aws-sdk-go upgrades for
+the latest controller release
+
+###### Controller Build
+* In the periodic job mentioned above, the ENTRYPOINT script will first
+validate whether there is a new minor version upgrade release for aws-sdk-go
```
RedbackThomson, line 106: Regarding my comment on filtering above, I think this is a bit too late to check for minor vs. patch. Probably better to just check asap.

docs/design/proposal/dependency-upgrades/aws-sdk-upgrades.md
```diff
@@ -0,0 +1,152 @@
...
+> NOTE: ACK runtime does not need to be updated for every aws-sdk-go
+> patch release since only *SDK Feature* updates and IAM service changes
+> impact ACK runtime functionality
```
RedbackThomson, line 109: nit: Unnecessary note.

docs/design/proposal/dependency-upgrades/aws-sdk-upgrades.md
```diff
@@ -0,0 +1,152 @@
...
+prowjob. As the number of service controllers increase, the result can be
+condensed into a separate summary file and individual results can be exported
+into separate files per controller and stored in a S3 bucket.
+* For every aws-sdk-go release upgrade, There will be auto created GitHub issue
+to track the actual release and these artifacts will be linked in that issue.
+* In the future, if needed, we can have a static webpage that can help service teams
+and external contributors browse these artifacts before the upgrade.
+
+### Release
+With the automation is place for **auto-generating** service controllers, this step can
```
RedbackThomson, line 132:
```suggestion
With the automation in place for **auto-generating** service controllers, this step can
```

docs/design/proposal/dependency-upgrades/aws-sdk-upgrades.md
```diff
@@ -0,0 +1,152 @@
...
+
+In future enhancements, the Verification step will also generate the Pull Request with
+new aws-sdk-go update for ACK runtime. Creating new ACK runtime patch release, updating
+code-generator and creating new code-generator release will also be automated when runtime
+PR gets merged.
+
+
+### Monitoring
+* Since this is one of the important jobs, when this job fails ACK core team should be
+updated by some mechanism.
```
RedbackThomson, line 148: What do you mean by "this is one of the most important jobs"? Why is checking for minor updates of the SDK so important?

RedbackThomson (changes requested):
Some minor nits, mostly. But I think this is practically ready.

docs/design/proposal/dependency-upgrades/aws-sdk-upgrades.md
vijtrip2: > "this is one of the ~most~ important jobs"

This seemed important to me because this is something we never pay attention to and can easily be missed if the prow job keeps failing. 
To your point, there is nothing very important with minor-version upgrade, it's a nice to have feature.

RedbackThomson (approved):
Cool. Nice design.
/lgtm

jaypipes (approved):
Nice proposal, @vijtrip2!

vijtrip2:
Updated the language. Thanks @jaypipes

PR #182, opened 2022-02-05
Automate soak bootstrapping with bash script

RedbackThomson:
Description of changes:
Creating a soak test AWS account, creating the cluster and bootstrapping all of the services into that cluster required an [extensive process](https://github.com/aws-controllers-k8s/test-infra/tree/main/soak). This process was well defined, but wordy and time consuming. 

This pull request replaces that entire process with a single bash script, `./bootstrap.sh`, which bootstraps an AWS account (creating an ECR public repository and an EKS cluster) and installs the soak testing manifests and charts onto the cluster. This will greatly reduce the number of steps required by service teams to onboard their soak test cluster. 

The [soak test installations instructions](https://github.com/aws-controllers-k8s/test-infra/tree/main/soak) have been updated to use the new script. The [soak test prow instructions](https://github.com/aws-controllers-k8s/test-infra/blob/93470a668d1870ef572824d0239a0c6570174fd6/soak/prow/README.md) have also been condensed into the main readme. 

Example of running the soak bootstrap script:
```
$ ./bootstrap s3 v0.0.12
```


soak/bootstrap.sh
```diff
@@ -0,0 +1,213 @@
...
+    echo "ok."
+fi
+
+# Install the controller into the cluster
+CONTROLLER_SERVICE_ACCOUNT_NAME=$(yq eval -e ".iam.serviceAccounts[0].metadata.name" $CLUSTER_CONFIG_PATH 2> /dev/null) || \
+    { >&2 echo "IRSA is not included in this cluster config. IRSA is required for soak testing"; exit 1; }
+
+ret=0
+helm list -n $CONTROLLER_NAMESPACE 2> /dev/null | grep -q $CONTROLLER_CHART_RELEASE_NAME || ret=$?
+if [ $ret -eq 0 ]; then
```
vijtrip2, line 140: We should uninstall the old chart with same name. Consider the situation where soak test was performed for ack-s3-controller `v0.10.0` and the controller-release helm chart was not uninstalled.

While reusing clear and testing `v0.11.0` , the past release should be upgraded or uninstalled.
I would prefer uninstall to get rid of dangling resources (if any)

soak/bootstrap.sh
```diff
@@ -0,0 +1,213 @@
...
+    echo "ok."
+fi
+
+# Final messages and links
+echo "Your soak test cluster should now be operational and actively running tests"
+echo "To port-forward your Grafana dashboard use the following command:"
+echo "    kubectl port-forward -n $PROM_NAMESPACE service/$PROM_CHART_RELEASE_NAME-grafana $LOCAL_GRAFANA_PORT:80 --address='0.0.0.0' >/dev/null &"
+echo "Then navigate to http://localhost:$LOCAL_GRAFANA_PORT/"
+echo ""
+echo "The default username is \"admin\" and the default password is \"prom-operator\""
```
vijtrip2, line 213: I believe a `soak/cleanup.sh` script is needed as well.

soak/README.md
```diff
@@ -1,317 +1,117 @@
...
+account.
+
+Export the credentials for an AWS user with administrator access to your
+terminal before following the rest of the documents.
+
+### Step 1 (Run the bootstrapping script)
+
+Run the `bootstrap.sh` script in the current directory. This script will do the
+following: 
+1. Create an ECR public repository for the soak test runner
```
vijtrip2, line 52: the ordered list only has "1" number

soak/README.md
```diff
@@ -1,317 +1,117 @@
...
+ACK-specific request counts and error codes used by the controller when making
+calls to AWS APIs. 
+
+The `Kubernetes/Compute Resources/Pod` dashboard will show the resource
+consumption by the controller pod.
+
+### Step 3 (Notify the ACK core team)
+
+The ACK core team needs to be made aware that service controller is ready for
+soak test execution using Prow, which is a manual process. Therefore, once you
```
vijtrip2, line 93: TL;DR: I am in favor of supporting a separate soak-test account to run soak-tests from prow but it may not work straight away as mentioned in this doc.

-----------

Soak tests on prow currently have a dependency that EKS cluster should be created in the same account as e2e tests. See current onboarding instructions [here](https://github.com/aws-controllers-k8s/test-infra/tree/main/soak/prow) . It will need a small amount of work to decouple that. 

Currently, ACK post-submit-job assumes the e2e-test-iam-role(from ssm) before executing `eks update-kubeconfig` command for gaining access to customer's  EKS soak cluster. Also we ask customers to create their EKS soak cluster using `e2e-test-iam-role` because this role will by default be authenticated by api-server, otherwise we get into whole spaghetti of `aws-auth` config map for authentication.

In the new solution, we will create another IAM role inside `soak/bootstrap.sh`, which will be used for running `eksctl create cluster` command and then we will store this IAM role in the SSM(to gain access to soak cluster) and the IRSA role(for installing ACK controller on soak cluster).

This week, I will write more documentation inside test-infra to describe this behavior and make above mentioned changes as well. But please remove this Step3 for now.

soak/README.md
```diff
@@ -1,317 +1,117 @@
...
+> **Note for Core Contributors:** Upon receiving a new IRSA ARN, access the
+ACK infrastructure account and add a new SSM string parameter with the path
+`/ack/prow/soak/irsa/<service>` and a value of the ARN.
+```bash
+# For ACK core contributors
+aws ssm put-parameter --name "/ack/prow/soak/irsa/$SERVICE" --type String
+--value <provided-value> 
+```
+
+### Step 4 (Customise your soak tests)
```
vijtrip2, line 109: I do not think `bootstrap.sh` consumes the `soak_config.yaml` file so this step is not valid.

vijtrip2 (changes requested):
Good work!

soak/README.md
RedbackThomson: With markdown you can create an ordered list by repeating `1.` and the renderer should fix the sequencing for you. At least Github does.

soak/README.md
RedbackThomson: Removing step 3 for now. I wonder if we could find some way to reduce the complexity here - by creating IAM roles with a standard name and using a service principal trust relationships - so we don't need to store stuff in SSM or mess with annoying kubeconfig messes.

soak/bootstrap.sh
RedbackThomson: Apart from deleting the cluster, what else would that achieve? I imagine we want to keep the cluster running, for future soak tests?

jaypipes (approved):
Fantastic improvement, @RedbackThomson, thank you! @vijtrip2 please feel free to send this on if you're OK with it.

soak/bootstrap.sh
vijtrip2: * uninstall soak-test-runner chart
* uninstall the controller-chart
* cleanup the port-forward bg processes if any.

above 3 will allow the cluster to be reused without old installations coming in the way.

soak/bootstrap.sh
```diff
@@ -0,0 +1,204 @@
...
+# Platform for soak-test-runner image.
+SOAK_IMAGE_PLATFORM="linux/amd64"
+# Release name of soak-test-runner helm chart.
+SOAK_RUNNER_CHART_RELEASE_NAME="soak-test-runner"
+
+# Total test duration is calculated as sum of TEST_DURATION_MINUTES,
+# TEST_DURATION_HOURS and TEST_DURATION_DAYS after converting them in
+# minutes. Override following variables accordingly to set your soak test
+# duration. Default value: 24 hrs.
+TEST_DURATION_DAYS=${TEST_DURATION_DAYS:-1}
```
vijtrip2, line 98: nit: Add these TEST_DURATION env variables in script usage

soak/bootstrap.sh
```diff
@@ -0,0 +1,204 @@
...
+fi
+
+$AWS_CLI eks update-kubeconfig --name $CLUSTER_NAME > /dev/null
+
+# Install the controller into the cluster
+CONTROLLER_SERVICE_ACCOUNT_NAME=$(yq eval -e ".iam.serviceAccounts[0].metadata.name" $CLUSTER_CONFIG_PATH 2> /dev/null) || \
+    { >&2 echo "Error: IRSA is not included in this cluster config. IRSA is required for soak testing"; exit 1; }
+
+if helm list -n $CONTROLLER_NAMESPACE 2> /dev/null | grep -q $CONTROLLER_CHART_RELEASE_NAME; then
+    echo "Controller Helm release ($CONTROLLER_CHART_RELEASE_NAME) already installed in cluster"
```
vijtrip2, line 136: shouldn't the installation be upgraded with latest controller image when it already exists?

vijtrip2 (approved):
Almost there!

soak/bootstrap.sh
RedbackThomson: What about I just focus on writing the script in such a way that we can re-use it for each release? Upgrading charts where applicable?

Also port-forwarding is not covered inside the script because ideally we would set up an ingress and put it behind a load balancer and domain name, so that it can be accessed by anyone. But that's for future us

soak/bootstrap.sh
RedbackThomson: Yeah. When I initially designed this script, I was thinking about how it works only for the first installation. I guess I should expand it to be used for any future releases, as well. I'll update this to match the soak test runner.

vijtrip2 (commented):
/lgtm

PR #184, opened 2022-02-28
Fix soak test service account name

RedbackThomson:
Description of changes:
We have to use a static service account name to put it inside the `eksctl` config file. Therefore, rather than referencing the default service account name, I have updated the soak test Helm chart to use the static one.


vijtrip2 (commented):
/lgtm

PR #187, opened 2022-03-11
Fix soak tests deploy region and add an env var to set controller repository URL

A-Hilaly:
Issue #, if available:

Description of changes:

- Use `DEPLOY_REGION` to set eksctl --region flag
- Add new environment variable to set controller image
- Properly propagate controller image tag for controller helm chart


RedbackThomson:
I quite purposefully didn't allow for changing the EKS cluster region, so that all soak tests clusters are in us-west-2. I have a feeling that when connecting it to Prow, if we have to switch regions for every service it will be a pain to manage them. It's easier to just use `aws eks update-kubeconfig --name ack-soak-test` with credentials from every account.

soak/bootstrap.sh
```diff
@@ -57,8 +59,10 @@ CONTROLLER_DIR="$ROOT_DIR/../$AWS_SERVICE-controller"
 CLUSTER_CONFIG_PATH="$SOAK_DIR/cluster-config.yaml"
 # EKS cluster name if not specified in cluster config
 DEFAULT_CLUSTER_NAME="ack-soak-test"
-# Semver version of the controller under test. Ex: v0.0.2
-CONTROLLER_CHART_URL="public.ecr.aws/aws-controllers-k8s/$AWS_SERVICE-chart"
+# Default controller image repository URL
+DEFAULT_CONTROLLER_IMAGE_REPO="public.ecr.aws/aws-controllers-k8s/$AWS_SERVICE"
```
vijtrip2, line 63: "-controller" suffix is missing. It should be "$AWS_SERVICE-controller" right?

A-Hilaly:
> I quite purposefully didn't allow for changing the EKS cluster region, so that all soak tests clusters are in us-west-2. I have a feeling that when connecting it to Prow, if we have to switch regions for every service it will be a pain to manage them. It's easier to just use aws eks update-kubeconfig --name ack-soak-test with credentials from every account.

Makes sense. From the `USAGE` var I thought that the DEPLOY_REGION would also set the region in which the eks cluster is created

PR #189, opened 2022-03-17
adding in OPERATOR_SDK_BIN_PATH env var to relevant OLM jobs and scripts

acornett21:
Issue #, if available:
- Relates: aws-controllers-k8s/code-generator#305 

Description of changes:
In order to get changes to code-generator made, the tests for OLM also need to be changed. Passing in the new env `OPERATOR_SDK_BIN_PATH` where necessary. Please let me know if I missed any places.

Signed-off-by: Adam D. Cornett <adc@redhat.com>

### DIFF
--- a/prow/jobs/images/Dockerfile.olm-bundle-pr
+++ b/prow/jobs/images/Dockerfile.olm-bundle-pr
@@ -10,6 +10,8 @@ ENV GOPATH=/home/prow/go \
     GO111MODULE=on \
     PATH=/home/prow/go/bin:/usr/local/go/bin:${PATH}
 
+ENV OPERATOR_SDK_BIN_PATH=/usr/local/bin
+
 RUN echo "Installing packages ..." \
     && apt-get update \
     && apt-get install -y --no-install-recommends\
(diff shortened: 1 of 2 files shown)


prow/jobs/jinja/presubmits/code_generator_tests.jinja2
```diff
@@ -38,6 +38,8 @@
           value: "s3"
         - name: ACK_GENERATE_OLM
           value: "true"
+        - name: OPERATOR_SDK_BIN_PATH
+          value: "/usr/local/bin"
```
RedbackThomson, line 42: I think I would prefer for this to be maintained inside the image. Since we are installing the binary as part of the Dockerfile, if we update that, we would also need to remember to update this Prow config. Would you mind just adding the export on a line underneath where we copy the operator SDK binary over? https://github.com/aws-controllers-k8s/test-infra/blob/main/prow/jobs/images/Dockerfile.olm-bundle-pr#L53-L54

cd/olm/olm-bundle-pr.sh
```diff
@@ -39,6 +39,7 @@ RELEASE_VERSION=$PULL_BASE_REF
 OLM_BUNDLE_VERSION=$(echo "$RELEASE_VERSION" | awk -F v '{print $NF}')
 echo "olm-bundle-pr.sh][INFO] olm bundle version is $OLM_BUNDLE_VERSION"
 export ACK_GENERATE_OLM=true
+export OPERATOR_SDK_BIN_PATH=/usr/local/bin
```
RedbackThomson, line 42: Echoing my comment below, but for this image too.

RedbackThomson (changes requested):

prow/jobs/jinja/presubmits/code_generator_tests.jinja2
acornett21: @RedbackThomson I moved this change to the docker files, and also made these two docker files match where appropriate. LMK if there are any other issues.

vijtrip2: I am not sure if exporting the variable in a Dockerfile layer will work make the env variable available during prow execution.


Why not add `ENV OPERATOR_SDK_BIN_PATH=/usr/local/bin` inside both Dockerfiles?

acornett21: @vijtrip2 I wasn't sure either, that's why I added it to the `jobs` first, but you're probably right then need to be ENV's as well.

vijtrip2:
/lgtm

I will build these new prow images, and publish them, and update the prow-jobs to use new image version, and then re-run the code-generator tests. :)

acornett21:
Thanks @vijtrip2!!

PR #196, opened 2022-03-31
Check code-gen version for auto-gen controllers

RedbackThomson:
Description of changes:
When auto-generating controllers, don't immediately skip the controller if the runtime version does not match. Instead, check if there is a difference in the ack-generate binary version as compared to the generated metadata file.


jaypipes (approved):
Nice. Changes make sense to me. Thanks @RedbackThomson

cd/auto-generate/auto-generate-controllers.sh
```diff
@@ -158,6 +160,16 @@ for CONTROLLER_NAME in $CONTROLLER_NAMES; do
       fi
       echo "ok"
     fi
+
+    SERVICE_AVAILABLE_API_VERSION=$(yq e '.api_versions[] | select(.status == "available") | .api_version' metadata.yaml)
```
vijtrip2, line 164: nit: add an empty check for "SERVICE_AVAILABLE_API_VERSION" value and skip if this is empty

vijtrip2 (approved):

PR #198, opened 2022-04-02
Convert ack-discover to Python package

RedbackThomson:
Description of changes:
In order to generate the service page for the documentation website, by being able to re-use the `ack-discover` CLI tool python methods, I converted it into a Python package with its own separate `setup.py` file. This can be installed independently from `acktest`, by installing it and using `import ackdiscover`.

I also made a few small changes to support the properties of controllers, releases and services we need for the website.


tools/ackdiscover/printer.py
```diff
@@ -11,15 +11,13 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-import dataclasses
+from dataclasses import dataclass, field, asdict
```
jaypipes, line 14: I dislike importing objects. Much prefer to stick with only importing modules... This makes it much more clear in the Python code following where a particular object is defined...

jaypipes (approved):
Excellent work on this, @RedbackThomson, thank you!

PR #208, opened 2022-05-12
Fluxv2 upgrade v0.13.1 -> v0.30.2

vijtrip2:
Description of changes:

* Update flux2 `GitRepository` and `Kustomization` manifests to v1beta2 as recommended by flux release notes
* Update the prow EKS cluster to k8s version v1_20, as the minimum requirement for flux2
* Failure to install fluxConfiguration was rolling back the flux installation as well in CloudFormation, so I refactored it out to bring back the prow cluster one component at a time.
* `package.lock.json` update caused by *npm install*

I will replicate the prow cluster locally and try to find a way which will allow flux upgrade without taking prow downtime, but this PR shall be merged to keep the cdk consistent with current state of cluster.

### DIFF
--- a/infra/lib/ci-cluster.ts
+++ b/infra/lib/ci-cluster.ts
@@ -4,11 +4,10 @@ import * as ec2 from '@aws-cdk/aws-ec2';
 import * as iam from '@aws-cdk/aws-iam';
 import * as cdk8s from 'cdk8s';
 import { policies as ALBPolicies } from './policies/aws-load-balancer-controller-policy';
-import { FluxChart } from './charts/flux';
 import { FluxConfigurationChart } from './charts/flux-configuration';
 import { ProwSecretsChart, ProwSecretsChartProps } from './charts/prow-secrets';
 import { NamespaceChart } from './charts/namespace';
-import { EXTERNAL_DNS_NAMESPACE, PROW_JOB_NAMESPACE, PROW_NAMESPACE } from './test-ci-stack';
+import { EXTERNAL_DNS_NAMESPACE, FLUX_NAMESPACE, PROW_JOB_NAMESPACE, PROW_NAMESPACE } from './test-ci-stack';
 
 export type CIClusterCompileTimeProps = ProwSecretsChartProps;
 
@@ -26,7 +25,7 @@ export class CICluster extends cdk.Construct {
     super(scope, id);
 
     this.testCluster = new eks.Cluster(scope, 'TestInfraCluster', {
-      version: eks.KubernetesVersion.V1_19,
+      version: eks.KubernetesVersion.V1_20,
       defaultCapacity: 0
     })
     this.testNodegroup = this.testCluster.addNodegroupCapacity('TestInfraNodegroup', {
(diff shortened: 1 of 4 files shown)


infra/lib/ci-cluster.ts
```diff
@@ -26,7 +25,7 @@ export class CICluster extends cdk.Construct {
     super(scope, id);
 
     this.testCluster = new eks.Cluster(scope, 'TestInfraCluster', {
-      version: eks.KubernetesVersion.V1_19,
+      version: eks.KubernetesVersion.V1_20,
```
jaypipes, line 28: was this necessary for the upgrade to Flux or were you just using this PR as an opportunity to bring our EKS cluster up to date?

infra/lib/ci-cluster.ts
```diff
@@ -49,17 +49,24 @@ export class CICluster extends cdk.Construct {
...
-      new FluxChart(
-        this.cdk8sApp, 'Flux', {}
-      )
-    );
+    const fluxNamespace = this.createNamespace(FLUX_NAMESPACE);
+    const fluxChart = this.testCluster.addHelmChart('flux2', {
+      chart: 'flux2',
+      repository: 'https://fluxcd-community.github.io/helm-charts',
+      namespace: FLUX_NAMESPACE,
+      version: '0.19.1',
```
jaypipes, line 57: maybe should be 0.19.2? https://github.com/fluxcd-community/helm-charts/blob/7fe7d95627b21a751ccd52725fe1a187cbada1ff/charts/flux2/Chart.yaml#L5

infra/lib/ci-cluster.ts
```diff
@@ -49,17 +49,24 @@ export class CICluster extends cdk.Construct {
   }
 
   installFlux = () => {
-    const fluxChart = this.testCluster.addCdk8sChart('flux',
-      new FluxChart(
-        this.cdk8sApp, 'Flux', {}
-      )
-    );
+    const fluxNamespace = this.createNamespace(FLUX_NAMESPACE);
+    const fluxChart = this.testCluster.addHelmChart('flux2', {
```
RedbackThomson, line 53: Thankfully, now we are using Helm, we can use `createNamespace: true` and not have to create the namespace as a dependency. This will also help with if we need to upgrade, it won't try to create the namespace each time.

RedbackThomson (changes requested):
Cool. This should hold us over for a little bit longer, until we need to switch off K8s 1.20

infra/lib/ci-cluster.ts
vijtrip2: It was necessary. Flux installation fails with a pre-check if the cluster is below k8s 1.20.6

infra/lib/ci-cluster.ts
vijtrip2: Thanks! I will change it.
Updating it may potentially take out all prow components again, so I will set out some time tomorrow(Friday) to do that.

vijtrip2: Done.

infra/lib/ci-cluster.ts
vijtrip2: Done!

PR #209, opened 2022-05-13
Add support for S3Bucket policy

A-Hilaly:
Issue #, if available:

Description of changes:
To correctly execute cloudtrail-controller e2e we will need to set the
trai bucket policy. This patch adds possibility of creating a bucket and
setting it policy directly with the test-infra e2e test framework.


jaypipes (approved):
++ seems reasonable to me, thank you @A-Hilaly :)

PR #213, opened 2022-05-24
Support more policy value replacements for Bucket

A-Hilaly:
Issue #, if available:

Currently, cloudtrail-controller uses the boostrappable `Bucket` object
to setup the Trail logs bucket. This bucket needs a special policy that
references the bucket name, trail name, region, and account ID. This
patch adds a new parameter (`policy_vars`) to `Bucket` that will help
users to replace more variables that the ones known by the acktest
library.

By submitting this pull request, I confirm that my contribution is made
under the terms of the Apache 2.0 license.

### DIFF
--- a/src/acktest/bootstrapping/s3.py
+++ b/src/acktest/bootstrapping/s3.py
@@ -13,6 +13,7 @@ class Bucket(Bootstrappable):
     name_prefix: str
     enable_versioning: bool = False
     policy: str = ""
+    policy_vars: dict = {}
 
     # Outputs
     name: str = field(init=False)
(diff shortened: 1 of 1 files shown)


RedbackThomson:
I don't think I quite understand the use case here. Is it not possible for these values to be defined when passing the `policy` string? Why do they need to be replaced inside this `Bootstrappable`?

A-Hilaly:
> I don't think I quite understand the use case here. Is it not possible for these values to be defined when passing the policy string? Why do they need to be replaced inside this Bootstrappable?

It is not possible to replace the bucket name before bootstrapping the bucket, [since the name is generated within the `bootstrap` function.](https://github.com/aws-controllers-k8s/test-infra/blob/main/src/acktest/bootstrapping/s3.py#L31)

RedbackThomson:
> It is not possible to replace the bucket name before bootstrapping the bucket, [since the name is generated within the `bootstrap` function.](https://github.com/aws-controllers-k8s/test-infra/blob/main/src/acktest/bootstrapping/s3.py#L31)

I see. Perhaps for that I can see this being a good exception, but otherwise I think I'd prefer the rest to remain hardcoded in policy. It gets tricky to understand what's going on with those policy documents if there are lots of substitutions happening. Nevertheless, these are testing tools so it's only exposed to developers, so maybe I'm being hypercritical.

A-Hilaly:
> It gets tricky to understand what's going on with those policy documents if there are lots of substitutions happening.

I get your concern, however, I believe that it's trickier to have replacements happening in two different and separate places... it might be confusing for developers to see that some values are getting magically replaced and others are not.

jaypipes (approved):
Yeah, I'm good with this. Would be great in the future to have some docstrings for properties like this, too. :)

PR #220, opened 2022-06-13
Add tag utils to acktest

RedbackThomson:
Description of changes:
Adds a new utility file for transforming ACK resource tag lists

### DIFF
--- a/src/acktest/tags.py
+++ b/src/acktest/tags.py
@@ -1,0 +1,25 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may
+# not use this file except in compliance with the License. A copy of the
+# License is located at
+#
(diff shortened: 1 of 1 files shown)


src/acktest/tags.py
```diff
@@ -0,0 +1,25 @@
...
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+"""Utilities for working with tags"""
+
+from typing import List
+
+ACK_SYSTEM_TAG_PREFIX = "services.k8s.aws/"
+
+def clean(tags: List[object]) -> List[object]:
```
vijtrip2, line 20: this is good but may need another method for handling tags which are map[string]string.
Or update this method to check for the type.

Happy to merge this though.

RedbackThomson: Yeah, I thought about that too. I'd probably rather we update this method to be a bit smarter, rather than having two distinct ones. But that's for future us to figure out

PR #222, opened 2022-06-28
Refactor e2e test scripts for declarative config

RedbackThomson:
Part of https://github.com/aws-controllers-k8s/community/issues/1359

Description of changes:
The existing e2e test scripts have organically grown into a mess of tangled environment variables and hardcoded path values. They do not allow for flexibility in how to run tests - you cannot run a subset of the tests, you can't run the tests on the local machine (only within Docker), you can't change the region for where the tests should run. 

This pull request moves the configuration of the tests into a declarative config file (`test_config.yaml`) which will live in the root of the `test-infra` directory on each developer's machine. This file will be ignored from git, as it contains AWS account IDs and possibly other private information, but an example configuration has been provided (`test_config.example.yaml`). As well, this pull request refactors each of the scripts into modular bash functions. Rather than using environment variables to pass information between scripts, parameters are passed as function arguments. Lastly, each script has been redesigned so that it can be imported independently from any other - each defines its own required binaries and inputs. As a result, developers can write custom scripts using these as the basis for their own building blocks, if they need to support other conditions.

Note:
This pull request removes the `install-*.sh` scripts. These scripts required maintenance to keep them at the latest version and typically only worked for amd64 Linux systems. Rather than converting these to be architecture-agnostic, I have just removed them and will update the docs to link to the CLI download pages instead.


scripts/lib/logging.sh
```diff
@@ -0,0 +1,50 @@
...
+# missing, we look for the existence of an environs variable called
+# "indent_level" and use that.
+_indented_msg() {
+    local __msg=${1:-}
+    local __indent_level=${2:-}
+    __indent=""
+    if [ -n "$__indent_level" ]; then
+        __indent="$( for each in $( seq 0 $__indent_level ); do printf " "; done )"
+    fi
+    echo "$__indent$__msg"
```
RedbackThomson, line 17: Note to self: Add timestamps

scripts/test-helm.sh
```diff
@@ -1,164 +0,0 @@
-#!/usr/bin/env bash
```
vijtrip2, line 1: substitute for this file?(used in prow tests)

scripts/test-aws-creds.md
```diff
@@ -1,82 +0,0 @@
-### Introduction
```
vijtrip2, line 1: new place for this file?

scripts/lib/gh.sh
```diff
@@ -1,113 +0,0 @@
-#!/usr/bin/env bash
-
```
vijtrip2, line 2: substitute for this file?(used in auto-generate-controllers prow job)

vijtrip2 (commented):
Nice refactoring! 🙌  Left some comments.

.gitignore
```diff
@@ -129,7 +129,7 @@ dmypy.json
 # Pyre type checker
 .pyre/
 
-# test generated file(s)
-scripts/web-identity-token
```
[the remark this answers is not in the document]
RedbackThomson: I actually modified the Dockerfile so that it only adds the `web-identity-token` if it exists. So we no longer need to create a temporary file

scripts/kind-setup.sh
```diff
@@ -0,0 +1,116 @@
...
+
+    local config_file_name="$(get_cluster_configuration_file_name)"
+    local cluster_version="$(get_cluster_k8s_version)"
+
+    local config_file_path=$SCRIPTS_DIR/kind-configurations/$config_file_name
+    
+    info_msg "Using configuration \"$config_file_name\""
+    debug_msg "Using K8s version \"$cluster_version\""
+
+    for i in $(seq 0 3); do
```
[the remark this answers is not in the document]
RedbackThomson: They come from the original script. I reduced it from 5 down to 3 retries. Sometimes KIND fails to bring up a cluster (I've had this happen many times). Usually the retries don't fix this, though, but I figured it was worth a shot in case it happens in Prow

scripts/kind-setup.sh
```diff
@@ -0,0 +1,116 @@
...
+_install_additional_controller() {
+    local __controller_service=$1
+    local __controller_version=$2
+    local __region=$3
+    local __controller_namespace=$4
+
+    debug_msg "Logging into the Helm registry"
+    _perform_helm_login
+
+    helm install --create-namespace -n "$__controller_namespace" \
```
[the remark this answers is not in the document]
RedbackThomson: Hmm actually I thought about this as I was pushing the code. We are installing the controllers but without any credentials, so they're essentially useless. We can't use IRSA because we can't assume they're running on EKS. However, we also can't use the `assumed_role_arn` because that might not have the IAM policies to run these controllers. Any ideas?

scripts/test-helm.sh
RedbackThomson: Just added. Good callout

scripts/lib/gh.sh
RedbackThomson: Should this be moved to the code-generator `scripts` directory instead?

scripts/test-aws-creds.md
RedbackThomson: I will rewrite parts of this file and bring it back into the directory. Good call out

scripts/lib/gh.sh
vijtrip2: It is mainly used by the CD process(auto-generating controllers). I think it should stay in test-infra.

scripts/kind-setup.sh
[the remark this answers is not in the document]
vijtrip2: I would say that we use the same assumed_role_arn, and permission of those role need to be extended when new cross-api references are added.

Another option would be managing multiple test iam roles for each AWS service (depending on which other controllers are installed), but i can see it getting messier to manage.

.gitignore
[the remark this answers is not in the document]
vijtrip2: very nice!

jaypipes (changes requested):
I really love this direction. A++

Couple comments inline but overall I think this is an awesome improvement. Thank you @RedbackThomson!

scripts/iam-policy-test-runner.sh
```diff
@@ -0,0 +1,62 @@
...
+            exit 1
+        fi
+    done
+}
+
+ensure_inputs() {
+    [[ -z "$AWS_SERVICE" ]] && { error_msg "Expected \`AWS_SERVICE\` to be defined"; exit 1; } || :
+}
+
+ensure_inputs
```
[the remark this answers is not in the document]
RedbackThomson: Actually yes. I worry that the scripts we provided don't install properly for all systems - namely any ARM64 system - and some may assume Darwin or Linux. Our documentation for the testing section has links for each of the dependencies, as well. Do we want to keep those install scripts up to date?

scripts/kind-setup.sh
```diff
@@ -38,8 +38,8 @@ _create_kind_cluster() {
     debug_msg "Using K8s version \"$cluster_version\""
 
     for i in $(seq 0 3); do
-        if [[ -z $(kind get clusters 2>/dev/null | grep $cluster_name) ]]; then
-            kind create cluster --name "$cluster_name" \
+        if [[ -z $(kind get clusters 2>/dev/null | grep $__cluster_name) ]]; then
```
vijtrip2, line 41:
```suggestion
        if [[ -z $(kind get clusters 2>/dev/null | grep "$__cluster_name") ]]; then
```

prow/jobs/test_config.yaml
```diff
@@ -0,0 +1,14 @@
+cluster:
+  create: true
+  k8s_version: 1.22.9
+
+aws:
+  region: us-west-2
+
+tests:
+  helm:
+    enabled: true
```
vijtrip2, line 10: is this still needed?

vijtrip2 (commented):
very nice! Two minor comments.

prow/jobs/test_config.yaml
RedbackThomson: Good call, thank you.

scripts/lib/gh.sh
RedbackThomson: Moved to `cd/lib` and anything using it has been updating accordingly

vijtrip2:
/approve

@jaypipes , feel free to merge this PR after your review.

Makefile
```diff
@@ -13,15 +13,11 @@ build-prow-jobs: ## Compiles the Prow jobs
...
-kind-test: export LOCAL_MODULES = false
-kind-test: ## Run functional tests for SERVICE with ACK_ROLE_ARN
-	@./scripts/kind-build-test.sh $(AWS_SERVICE)
-
-local-kind-test: export PRESERVE = true
-local-kind-test: export LOCAL_MODULES = true
-local-kind-test: ## Run functional tests for SERVICE with ACK_ROLE_ARN allowing local modules
-	@./scripts/kind-build-test.sh $(AWS_SERVICE)
+kind-test: ## Run functional tests for SERVICE
+	@AWS_SERVICE=$(AWS_SERVICE) ./scripts/start.sh
```
jaypipes, line 17: Can we call this script `scripts/run-e2e-tests.sh` please?

scripts/controller-setup.sh
```diff
@@ -0,0 +1,181 @@
...
+source "$SCRIPTS_DIR/lib/config.sh"
+source "$SCRIPTS_DIR/lib/logging.sh"
+
+build_and_install_controller() {
+    local __cluster_name=$1
+    local __controller_namespace=$2
+    local __img_name=$3    
+
+    info_msg "Building controller image ... "
+    _build_controller_image $__img_name
```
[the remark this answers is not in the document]
jaypipes: @RedbackThomson did you want to update the code in the Bash files to enclose parameters in double quotes?

scripts/iam-policy-test-runner.sh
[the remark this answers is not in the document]
jaypipes: @RedbackThomson OK, fair enough. Maybe add a note to the PR summary that we're removing the install scripts because they were not platform-agnostic and too difficult to keep up to date?

scripts/pytest-image.Dockerfile
```diff
@@ -13,11 +13,15 @@ ARG TEST_INFRA_PATH=./test-infra
 # Destination path to copy aws-web-identity-token
```
jaypipes, line 13: nit: the other Dockerfiles in [`prow/jobs/images`](https://github.com/aws-controllers-k8s/test-infra/tree/main/prow/jobs/images) all use the form `Dockerfile.$suffix` instead of `$prefix.Dockerfile`

scripts/start-helm-tests.sh
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
```
jaypipes, line 1: consider renaming this file `run-helm-tests.sh`, moving the `install_chart_and_run_tests` method from `helm-test-runner.sh` to this file, and renaming `helm-test-runner.sh` to just `helm.sh`

scripts/helm-test-runner.sh
```diff
@@ -0,0 +1,119 @@
...
+
+HELM_DIR="$SERVICE_CONTROLLER_SOURCE_PATH/helm"
+
+source "$SCRIPTS_DIR/lib/common.sh"
+source "$SCRIPTS_DIR/lib/config.sh"
+source "$SCRIPTS_DIR/lib/logging.sh"
+
+source "$SCRIPTS_DIR/controller-setup.sh"
+
+install_chart_and_run_tests() {
```
jaypipes, line 21: This feels a bit oddly coupled, to be honest. I think there should be totally separate scripts that do installation/setup and scripts that have tests in them. Recommend moving `install_chart_and_runt_tests` to the `run-helm-tests.sh` script below and renaming this file to simply `scripts/helm.sh`

scripts/kind-setup.sh
```diff
@@ -0,0 +1,116 @@
+#!/usr/bin/env bash
```
jaypipes, line 1: consider naming this file simply `scripts/kind.sh`

jaypipes (changes requested):
more comments...

PR #234, opened 2022-07-20
Fix Prow web identity file passing

RedbackThomson:
Description of changes:
Fixes the authentication issues for the Prow cluster's test scripts. As well as a bug in the container credentials file, the file was not properly being passed to the Pytest container underneath. This PR was tested against the production Prow cluster to ensure it works for our IRSA setup.

Note: I have also removed the Prowjob's CPU requests, since there is no need for them to access 2 vCPUs worth of resources for every test run. This is unrelated to the web identity file work.


scripts/pytest-image-runner.sh
```diff
@@ -34,12 +36,11 @@ build_pytest_image() {
     local __image_tag=$1
 
     local ack_role_arn=$(get_assumed_role_arn)
```
jaypipes, line 38: have I mentioned it is confusing to have an AWS_ROLE_ARN and an ACK_ROLE_ARN? :)

scripts/pytest-image-runner.sh
```diff
@@ -50,7 +51,7 @@ build_pytest_image() {
         # NOTE: credentials in both these profiles rotate automatically
 
         # copy web-identity-token file for use inside test container
-        cp "$AWS_WEB_IDENTITY_TOKEN_FILE" "$SCRIPTS_DIR"/web-identity-token >/dev/null
+        cp "$identity_file" "$SCRIPTS_DIR"/web-identity-token >/dev/null
```
jaypipes, line 54: Above, you are defining TEST_CONTAINER_WEB_IDENTITY_TOKEN_FILE hard-coded to `/root/web-identity-token`. Any reason this is not using `$TEST_CONTAINER_WEB_IDENTITY_TOKEN_FILE` as the destination location for the copy?

scripts/pytest-image-runner.sh
RedbackThomson: I totally agree. `AWS_ROLE_ARN` is supplied to the test container by IRSA. `ACK_ROLE_ARN` is an unfortunately named variable that I have only kept for the sake of backward compatibility. I'd be happy to change it if we're willing to break that.

scripts/pytest-image-runner.sh
RedbackThomson: That env var defines the location of the identity token file within the Pytest container. This line copies the identity file from its location within the Prow container into the temporary directory we use when building the Pytest container. I know that's a bit confusing, but essentially we have to keep track of the file on the host (the Prow container) and the container (the Pytest container) both during build- and then run-time.

jaypipes: OK, makes sense. Might be good to add the above as a comment in the file somewhere :)

scripts/pytest-image-runner.sh
RedbackThomson: Give me 10 minutes

RedbackThomson: Updated and also created https://github.com/aws-controllers-k8s/community/pull/1399

jaypipes (approved):
:+1:

PR #236, opened 2022-07-21
Add trap for rotating AWS credential sleep

RedbackThomson:
Description of changes:
In the Prow cluster, the `sleep` command was becoming a hanging process and causing the container to never exit. On my development machine, and inside a `debian-slim` container, I was not able to reproduce this since killing the calling method (in this case `_loop_rotate_temp_creds`) killed all child sleep processes. This pull request adds a SIGINT and EXIT trap for the rotating credentials loop, and a trap inside the rotating loop function for the `sleep`, to ensure that any forked processes are killed every time.

### DIFF
--- a/scripts/run-e2e-tests.sh
+++ b/scripts/run-e2e-tests.sh
@@ -95,6 +95,7 @@ run() {
     ensure_cluster
     
     build_and_run_tests
+    exit $?
 }
 
 ensure_inputs() {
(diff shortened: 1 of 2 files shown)


scripts/run-e2e-tests.sh
```diff
@@ -95,6 +95,7 @@ run() {
     ensure_cluster
     
     build_and_run_tests
+    exit $?
```
jaypipes, line 98: :+1: this seems like the most likely culprit for the success-on-failure thing we were seeing, right?

jaypipes (approved):
That's some funky trapping there, @RedbackThomson

PR #242, opened 2022-07-25
add perform_buildah_and_helm_login command

vijtrip2:
Description of changes:
* fixing the ACK controller release job, by adding `perform_buildah_and_helm_login` command
* The same function existed before the refactoring so there is no new logic
* There is a scope for better refactoring but currently i am just adding same function to unblock controller release


jaypipes (approved):
:+1:

PR #244, opened 2022-07-26
Update the `yq` version inside the test container

RedbackThomson:
Description of changes:

Updates the version of `yq` used inside the e2e test container so that we can pin it to a known good version to match the syntax we are using


vijtrip2 (commented):
/lgtm

PR #249, opened 2022-08-04
Add `SigningProfile` bootsrappable resource

A-Hilaly:
Description of changes:

This patch adds `SigningProfile` bootstrappable resource to acktest
library. Note that signing profiles cannot be deleted. They can only
be canceled during the clean-up phase. They are automatically
deleted after two years.

### DIFF
--- a/src/acktest/bootstrapping/signer.py
+++ b/src/acktest/bootstrapping/signer.py
@@ -1,0 +1,49 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may
+# not use this file except in compliance with the License. A copy of the
+# License is located at
+#
+#	 http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+import boto3
+
+from dataclasses import dataclass, field
+
+from . import Bootstrappable
+from .. import resources
+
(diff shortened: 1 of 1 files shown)


src/acktest/bootstrapping/dynamodb.py
```diff
@@ -29,7 +29,7 @@ def dynamodb_resource(self):
     def bootstrap(self):
         """Creates a Dynamodb table with an auto-generated name.
         """
-        self.name = resources.random_suffix_name(self.name_prefix, 63)
+        self.name = resources.random_suffix_name(self.name_prefix, 8)
```
vijtrip2: You can easily blow past 8 character limit with a table name from tests. Why did you need to change this?

vijtrip2 (approved):

src/acktest/bootstrapping/dynamodb.py
A-Hilaly: Modified the wrong file :) just force pushed

A-Hilaly: Signing profile names cannot go over 64characters

vijtrip2:
>They can only be canceled during the clean-up phase. They are automatically deleted after two years.

Argh... hopefully it does not cause e2e resources to pileup and then fail due to exceeded limits. But i guess we can do nothing about it.

src/acktest/bootstrapping/dynamodb.py
vijtrip2: why not make it something like 32? instead of 8.

IIRC, this method will only add suffix if name_prefix is less than 8 chars which is very less IMO. Consider 32 if you don't mind

A-Hilaly: Agreed

vijtrip2 (commented):
/lgtm

PR #258, opened 2022-11-04
Add more controllers to the list of e2e tests to be run by Prow 

jljaco:
...when code-generator changes happen

### DIFF
--- a/prow/jobs/jobs_config.yaml
+++ b/prow/jobs/jobs_config.yaml
@@ -45,4 +45,10 @@ soak_test_on_release_service_names: []
 code_gen_presubmit_services:
 - dynamodb
 - ecr
+- eks
+- iam
+- lambda
+- prometheusservice
+- rds
 - s3
+- sfn


jaypipes (commented):
@jljaco I'm not sure it's useful to e2e test all of these service controllers on each change to code-generattor and runtime packages. Is there a particular issue you were interested in solving by adding these service controllers to the list of e2e tests for code-gen/runtime?

prow/jobs/jobs_config.yaml
```diff
@@ -45,4 +45,10 @@ soak_test_on_release_service_names: []
 code_gen_presubmit_services:
 - dynamodb
 - ecr
+- eks
```
RedbackThomson, line 48: The EKS tests take >40 minutes to run. I'd be really hesitant to add this as it would block any fast changes to code-generator. Are you trying to validate a specific functionality in every code gen PR? What necessitated these changes?

jljaco:
Spoke with @jaypipes offline about this idea.  Was pointed to https://github.com/aws-controllers-k8s/test-infra/blob/main/cd/auto-generate/auto-generate-controllers.sh which actually accomplishes this is a much more reasonable way.

PR #261, opened 2022-11-17
Use `apiserver_storage_objects` as new soak metric for resource counts

jljaco:
Motivation:  noticed when running soak tests that the `ACK Resource Count` graph was blank.  After some digging, determined that the metric `etc_object_counts` is now deprecated upstream in favor of `apiserver_storage_objects`.

See:

- https://github.com/kubernetes/apiserver/blob/464f2d738348f8048b8791f4c6a7bf51144c1ac7/pkg/storage/etcd3/metrics/metrics.go#L50
- https://github.com/kubernetes/kubernetes/issues/98270 
- https://github.com/kubernetes/kubernetes/pull/99785 
- https://access.redhat.com/solutions/6969249

Before:
image

After:
image

Description of changes:
Change the default Grafana dashboard used to monitor ACK soak tests to use this new metric.

### DIFF
--- a/soak/monitoring/grafana/ack-dashboard-source.json
+++ b/soak/monitoring/grafana/ack-dashboard-source.json
@@ -533,7 +533,7 @@
         {
           "datasource": "$datasource",
           "exemplar": true,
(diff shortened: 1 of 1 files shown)


A-Hilaly (approved):
Great catch, thank you!
/lgtm

PR #263, opened 2022-11-21
Add `label_sync` periodic job

A-Hilaly:
This patch introduces a new prow periodic job that will be responsible
of syncing Github labels across all ACK repositories.

This job uses k/test-infra [`label_sync`](https://github.com/kubernetes/test-infra/tree/master/label_sync) tool.
`label_sync` consumes a labels configuration file and a github token to
sync labels in all the targeted repositories. 

PS: Setting the periodic job frequency to 6 hours for now (might increase
the value later)

Included in patch:
- New periodic job called `label-sync`
- New `label.yaml` generation scripts. This one will be responsible of
  bringing new labels whenever a new service is introduced.
- Mount `label.yaml` configuration file as a volume for the periodic
  job
- Update `prow/jobs/README.md` file


A-Hilaly:
Note to reviewers: You can find the list of labels that will be synced here https://github.com/aws-controllers-k8s/test-infra/blob/76190ad61bb6d51875786cf9f51ebb5686dd62cf/prow/jobs/labels.yaml

A-Hilaly:
I will need to execute the `label_sync` locally before sending this to prod.
/hold

RedbackThomson:
Can we generate the service labels using the jinja scripts we already have?

A-Hilaly:
>Can we generate the service labels using the jinja scripts we already have?

@RedbackThomson  i wanted to go for that solution initially, but thought that maybe we'll need to rename the script. Thoughts?

RedbackThomson:
> >Can we generate the service labels using the jinja scripts we already have?
> 
> 
> 
> @RedbackThomson  i wanted to go for that solution initially, but thought that maybe we'll need to rename the script. Thoughts?

Could we make it into a different `make` command? We can rename the python files as much as we like

jaypipes (approved):
I'm good with this. +1

prow/jobs/jobs.yaml
```diff
@@ -1,6 +1,41 @@
...
+      command:
+      - label_sync
+      args:
+      - --config=/etc/config/labels.yaml
+      - --confirm=true
+      - --orgs=aws-controllers-k8s
+      - --token=/etc/github/oauth
+      - --github-endpoint=http://ghproxy
+      - --github-endpoint=https://api.github.com
+      - --debug
```
jljaco, line 24: Do you want to keep this `--debug` here?

prow/jobs/jinja/periodics/label_sync.jinja2
```diff
@@ -0,0 +1,35 @@
...
+      command:
+      - label_sync
+      args:
+      - --config=/etc/config/labels.yaml
+      - --confirm=true
+      - --orgs=aws-controllers-k8s
+      - --token=/etc/github/oauth
+      - --github-endpoint=http://ghproxy
+      - --github-endpoint=https://api.github.com
+      - --debug
```
jljaco, line 21: Ditto, do you want this `--debug`?

prow/jobs/jobs.yaml
A-Hilaly: Left on purpose to see how to first job execution will go. Planning on removing this flag sometime in Q1.

PR #282, opened 2023-01-10
Offboard eventbridge from Prow jobs

RedbackThomson:
Description of changes:
The `eventbridge-controller` repository currently doesn't contain any `apis` directory. This is causing our `autogenerate-controllers` job to fail, as it requires the `ack-generate-metadata` file to compare version numbers.

This PR temporarily disables the `eventbridge-controller` in Prow - we can re-enable this once we have an API to merge for that service.


jaypipes (approved):
:+1:

PR #288, opened 2023-01-24
Publish v0.0.13 image to resolve `kind` binary not found issue (Part 2)

A-Hilaly:
In the process of debugging the issue with the `/usr/bin` binary not
being found, I published a `v0.0.12` without including the necessary
`mv` command to properly place the binaries in the correct location.
This resulted in one of the nodes in the Prow cluster pulling the
incorrect image during testing.

With the fix included in the previous PR, the second node was able to
pull the corrected image, however, the first node was not able to update
to the latest image version hash (which an expected behavior in
Kubernetes land).

This patch publishes a new image version `v0.0.13` to ensure that all
nodes in the Kubernetes cluster are pulling the same, corrected image to
resolve the `/usr/bin` bug.


RedbackThomson:
Thanks!
/lgtm

PR #292, opened 2023-01-27
Lint all shell scripts with shellcheck

RedbackThomson:
Description of changes:
Used the output of the `shellcheck` tool to make style changes to all of the shell scripts within the repository. Added a new `make` target for `lint-shell` which lints all of the scripts and excludes [SC1091](https://www.shellcheck.net/wiki/SC1091), which fails because we `source` other scripts using path variables (a known limitation of `shellcheck`), and [SC2015](https://www.shellcheck.net/wiki/SC2015), which fails because we are using ternaries for some part of the configuration checking code which `shellcheck` warns may cause early execution of functions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ PROW_JOBS_PATH="./prow/jobs"
 
 AWS_SERVICE=$(shell echo $(SERVICE) | tr '[:upper:]' '[:lower:]')
 
-.PHONY: build-prow-jobs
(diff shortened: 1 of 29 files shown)


A-Hilaly (approved):
Looks good to me, maybe we should rebuild our prow jobs to make sure everything works smoothly?

RedbackThomson:
Going to close this out since it's gone stale. Can always re-open and rebase this PR when we're ready to go full force into linting

PR #295, opened 2023-02-01
Regenerate `jobs.yaml` and `labels.yaml`

A-Hilaly:
Regenerate `jobs.yaml` and `labels.yaml`


prow/jobs/jobs.yaml
```diff
@@ -3285,7 +3285,7 @@ postsubmits:
           - name: GITHUB_TOKEN
             valueFrom:
               secretKeyRef:
-                name: prowjob-github-token
+                name: prowjob-github-pat-token
```
A-Hilaly, line 3288: Not sure if this is correct, @jljaco @RedbackThomson ?

prow/jobs/labels.yaml
```diff
@@ -525,6 +525,11 @@ repos:
       name: service/emrcontainers
       target: issues
       addedBy: anyone
+    - color: f59745
+      description: Indicates issues or PRs that are related to eventbridge-controller.
+      name: service/eventbridge
+      target: issues
+      addedBy: anyone
```
A-Hilaly, line 532: I forgot to regenerate labels.yaml my last PR to test-infra

prow/jobs/jobs.yaml
jljaco: I believe this is correct

https://github.com/aws-controllers-k8s/test-infra/blob/9b24d142619fd2058415065b7e3ec3a11900b892/infra/lib/charts/prow-secrets.ts#L51

jljaco (approved):

RedbackThomson:
I genuinely forget to generate labels each time I rebuild the jobs. I now see we have a `gen-all`, but could we also make the default `make` target just call `gen-all` so I don't have to remember?

Otherwise this just needs another rebase and we can merge

PR #296, opened 2023-02-01
[acktest] Convert the expected input to have the same format as the actual input.

A-Hilaly:
Changes:
This patch adds a new feature to `assert_equal` that will convert the expected
input to have the same format as the actual input. This is necessary since some
AWS APIs return tag object with capitalized Key and Value fields.

### DIFF
--- a/src/acktest/tags.py
+++ b/src/acktest/tags.py
@@ -108,6 +108,14 @@ def assert_equal(expected: Union[dict, list],
     """
     assert len(actual) == len(expected), \
         'length of "expected" and "actual" tags is not same'
+
+    # Convert the expected input to have the same format as the actual input.
+    # Some service APIs return a list of tag objects with `key` and `value` in lower case,
+    # however, some other APIs use capitalized `Key` and `Value` (CloudTrail, EventBridge APIs).
+    # In the latter case, using this function will not work properly without transforming the
+    # expected objects to be similar to the actual ones.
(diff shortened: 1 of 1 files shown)


jljaco (approved):
looks good, I would like to see this actually used to resolve a problem in a test somewhere; can you link in this PR's description to another PR that uses this?

A-Hilaly:
>looks good, I would like to see this actually used to resolve a problem in a test somewhere; can you link in this PR's description to another PR that uses this?

I don't have an open PR using this, yet. But will definitely be helpful to simplify code in cloudtrail and eventbridge

src/acktest/tags.py
```diff
@@ -108,6 +108,14 @@ def assert_equal(expected: Union[dict, list],
     """
     assert len(actual) == len(expected), \
         'length of "expected" and "actual" tags is not same'
+
+    # Convert the expected input to have the same format as the actual input.
+    # Some service APIs return a list of tag objects with `key` and `value` in lower case,
+    # however, some other APIs use capitalized `Key` and `Value` (CloudTrail, EventBridge APIs).
+    # In the latter case, using this function will not work properly without transforming the
+    # expected objects to be similar to the actual ones.
+    expected = {key_member_name: v for k, v in expected.iteritems()}
```
A-Hilaly, line 117: Actually... this will not work for arrays....

PR #299, opened 2023-02-06
Upgrade Prow ALB controller to `1.4.7`

RedbackThomson:
Description of changes:
After upgrading the Prow cluster to K8s `1.22`, the ALB controller went into `CrashLoopBackOff` because it required `betav1/Ingress` which was deprecated and removed in `1.22`. This fix upgrades the ALB controller to the latest version so that it can be restored.


A-Hilaly (approved):
/lgtm

PR #301, opened 2023-02-08
Upgrade to CDK 2 and EKS to 1.24

RedbackThomson:
Description of changes:
CDK 1 only supports EKS up to 1.21, which is now considered deprecated. In order to upgrade the cluster, I had to upgrade to CDK 2. Along with upgrading to 1.24, I had to upgrade the Prow control plane components that are now considered deprecated (namely Ingress `betav1` and CustomResourceDefinition `betav1`).

The new Prow CRD is 43k lines long, which is too big to be supported by Helm. To manage its installation, it is being included as a separate Kustomization, and then the Prow control plane, data plane and jobs are added to another Kustomization which has a dependency on the CRDs (so they are installed in the correct order). 

This system was tested by deploying it to my local AWS account, pointing the cluster to my private branch so that I could check that all of the Flux parts reconciled properly. 


infra/lib/ci-cluster.ts
```diff
@@ -15,26 +12,24 @@ import {
...
 
   readonly namespaceManifests: eks.KubernetesManifest[];
 
-  constructor(scope: cdk.Construct, id: string, props: CIClusterProps) {
+  constructor(scope: Construct, id: string, props: CIClusterProps) {
     super(scope, id);
 
     this.testCluster = new eks.Cluster(scope, "TestInfraCluster", {
-      version: eks.KubernetesVersion.V1_21,
+      version: eks.KubernetesVersion.V1_24,
```
A-Hilaly, line 32: Will CDK handle small steps of version-to-version upgrades? or did you make bumps one by one?

prow/config/templates/deck-Ingress.yaml
```diff
@@ -13,30 +13,36 @@
 # limitations under the License.
 
 {{ if .Values.ingress.create }}
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1
```
A-Hilaly, line 16: Nice to see this GA

A-Hilaly (approved):
Niice! Glad we're moving to cdk 2 - thank you!

infra/lib/ci-cluster.ts
RedbackThomson: I actually think a jump like this would not be possible if we were upgrading the cluster, but with our plan to recreate the cluster I don't think it'll matter

A-Hilaly:
THank you @RedbackThomson 
/lgtm

PR #302, opened 2023-02-09
Downgrade Prow to K8s `1.23` and switch to `eks-blueprints`

RedbackThomson:
Description of changes:
`dockershim` was removed in K8s `1.24`, and we rely on `docker` to run all of the integration tests in the cluster. For now we will need to downgrade to `1.23` to support the current set of tests. 

Our previous upgrade from `1.21` also broke the `ghproxy` deployment in the Prow Helm chart because it requires a `PersistentVolumeClaim`, which used to utilise the EKS built-in EBS CSI driver. As of `1.23`, EKS no longer provides the EBS CSI driver, meaning we now need to install this into the cluster to create PVCs backed by EBS.

Upgrading to CDK 2 also allows us to now use the `eks-blueprints` module to simplify and harden our CDK cluster configuration. It also lets us install `Karpenter` with one line of code!

The SSM parameters that we used to have in here was not being used - we removed the `DaemonSet` that we used for that a long time ago. It wasn't compatible with the EKS blueprints configuration anyway so I have stripped it here. 


infra/bin/test-ci.ts
```diff
@@ -34,13 +34,13 @@ new TestCIStack(app, "TestCIStack", {
...
+    hostedZoneId:
+      app.node.tryGetContext("hosted_zone_id") || process.env.HOSTED_ZONE_ID,
   },
   logsBucketName:
     app.node.tryGetContext("logs_bucket") || process.env.LOGS_BUCKET,
   logsBucketImport:
     app.node.tryGetContext("logs_bucket_import") ||
     process.env.LOGS_BUCKET_IMPORT ||
     false,
-  pvreBucketName:
```
jljaco, line 44: Is this no longer used anywhere?

infra/lib/ci-cluster.ts
```diff
@@ -1,58 +1,80 @@
...
+      .region(Stack.of(this).region)
+      .version(clusterVersion)
+      .clusterProvider(new blueprints.MngClusterProvider(mngProps))
+      .resourceProvider(
+        GlobalResources.HostedZone,
+        new ImportHostedZoneProvider(props.hostedZoneId)
+      )
+      .addOns(
+        new blueprints.addons.VpcCniAddOn(),
+        new blueprints.addons.KarpenterAddOn(),
```
jljaco, line 61: 🔥

infra/lib/ci-cluster.ts
```diff
@@ -1,58 +1,80 @@
...
+        new blueprints.addons.EbsCsiDriverAddOn(),
+        new blueprints.addons.ExternalDnsAddOn({
+          hostedZoneResources: [GlobalResources.HostedZone],
+        })
+      )
+      .build(this, "TestInfraCluster");
+
+    this.testCluster = blueprintStack.getClusterInfo().cluster;
+
+    this.namespaceManifests = [PROW_JOB_NAMESPACE, PROW_NAMESPACE].map(
```
jljaco, line 72: Do we not need to create `EXTERNAL_DNS_NAMESPACE` now?  Is that handled by the `ExternalDnsAddOn`?

infra/package-lock.json
```diff
@@ -1153,10 +1340,21 @@
...
+      },
+      "funding": {
+        "url": "https://github.com/sponsors/ljharb"
+      }
+    },
     "node_modules/aws-cdk-lib": {
-      "version": "2.63.2",
-      "resolved": "https://registry.npmjs.org/aws-cdk-lib/-/aws-cdk-lib-2.63.2.tgz",
-      "integrity": "sha512-Kk9aP79O+9UJ3Bw0wjquMD+RpWPQ/Fa4wUcP5mdpROpoipPXsFFZPTXT8j8kp2oFOZa9OHxacTF06+/M0H/bWQ==",
+      "version": "2.60.0",
```
jljaco, line 1355: Any particular reason why we need to downgrade `aws-cdk-lib` here?  Shouldn't we be using the latest?

infra/package.json
```diff
@@ -11,19 +11,20 @@
...
-    "constructs": "^10.0.0",
-    "aws-cdk-lib": "^2.0.0"
+    "aws-cdk-lib": "2.60.0",
+    "constructs": "^10.0.0"
   },
   "devDependencies": {
     "@types/jest": "^26.0.10",
     "@types/node": "10.17.27",
-    "aws-cdk-lib": "^2.0.0",
+    "aws-cdk-lib": "^2.60.0",
```
jljaco, line 20: IMO we should fix the version of `aws-cdk-lib` instead of using this ">=" style of dependency specification

jljaco (commented):
Thank you so much for doing this -- a few minor comments

infra/bin/test-ci.ts
RedbackThomson: The SSM parameters that we used to have in here was not being used - we removed the `DaemonSet` that we used for that a long time ago. It wasn't compatible with the EKS blueprints configuration anyway so I have stripped it here.

infra/lib/ci-cluster.ts
RedbackThomson: Handled by the addon now. In fact I had it in there before and after waiting 45 minutes for the stack to create it failed on the last step - which was creating that namespace. 😠

infra/package-lock.json
RedbackThomson: Unfortunately the latest version had incompatibilities with `eks-blueprints`. Weird stuff like the `ec2.InstanceType` wouldn't work. I had to downgrade to the version they list.

infra/package.json
RedbackThomson: I agree, but we are restricted by `eks-blueprints` (see above)

jljaco (approved):
/lgtm (just for fun since Prow isn't listening right now ;)

infra/lib/ci-cluster.ts
```diff
@@ -1,58 +1,80 @@
...
 } from "./test-ci-stack";
+import {
+  GlobalResources,
+  ImportHostedZoneProvider,
+} from "@aws-quickstart/eks-blueprints";
 
-export type CIClusterCompileTimeProps = ProwGitHubSecretsChartProps;
+export type CIClusterCompileTimeProps = ProwGitHubSecretsChartProps & {
+  hostedZoneId: string;
+};
```
jaypipes, line 21: huh.. TIL...

So the `&` operator can be used against a variable of type `object` to "append" properties to that object?

infra/lib/ci-cluster.ts
```diff
@@ -1,58 +1,80 @@
...
+      .clusterProvider(new blueprints.MngClusterProvider(mngProps))
+      .resourceProvider(
+        GlobalResources.HostedZone,
+        new ImportHostedZoneProvider(props.hostedZoneId)
+      )
+      .addOns(
+        new blueprints.addons.VpcCniAddOn(),
+        new blueprints.addons.KarpenterAddOn(),
+        new blueprints.addons.AwsLoadBalancerControllerAddOn(),
+        new blueprints.addons.EbsCsiDriverAddOn(),
```
jaypipes, line 63: I think it's kinda crazy that the EBS CSI driver is no longer installed by default in EKS clusters. In fact, that is pointed out in [this article](https://medium.com/@elliotgraebert/comparing-the-top-eight-managed-kubernetes-providers-2ae39662391b) specifically as causing major headaches compared to GKE and AKS...

jaypipes (commented):
@RedbackThomson thanks for doing this. I wanted to learn a bit more specifically about the changes to the DNS setup for our cluster. Can you write a couple sentences as to why those changes were needed? Thanks!

jljaco:
This was a follow-up to https://github.com/aws-controllers-k8s/test-infra/pull/301

infra/lib/ci-cluster.ts
RedbackThomson: Yes it's a union of the types. Typescript has some crazy type definition helpers

RedbackThomson:
> I wanted to learn a bit more specifically about the changes to the DNS setup for our cluster. Can you write a couple sentences as to why those changes were needed?

The only differences that this PR makes to the DNS setup is that we are now installing the `external-dns` Helm chart through the `eks-blueprints` package. Part of the constructor for its add-on is a list of `HostedZoneIDs` so it can allowlist those within the controller. I guess this is to prevent the controller from overriding HostedZones it shouldn't be able to access. Otherwise, the DNS situation is the same - we have a Route53 HostedZone that is set up independently of this CDK and we have an Ingress that sets up a load balancer. The `external-dns` controller glues those together by adding records to the HostedZone with the public address of the load balancer.

infra/lib/ci-cluster.ts
RedbackThomson: Oh `ProwGitHubSecretsChartProps` is also a `type`, though, not an `object`

PR #303, opened 2023-02-09
Fix Prow ingress priority

RedbackThomson:
Description of changes:
I noticed that the webhooks weren't working. When looking at the webhook panel within the Github app I saw that they were being met with 404s at `/hook`. Doing some digging, I found that the 404s were being returned by the `deck` service rather than `hook`. After adjusting the order that the ingress defines its routes, putting `/hook` and `/metrics` before `/*`, it gave precedence to the more specific subroutes and the webhooks were able to be delivered.

In this PR I also downgraded the `hook` from a `LoadBalancer` to a `ClusterIP` by default. Since this service is exposed through the ingress, which is itself behind a `LoadBalancer`, there is no need to serve it with an ALB. This is how it works for `deck` and so just to stay consistent this is how it should also work for `hook`.

### DIFF
--- a/prow/config/templates/ing-Ingress.yaml
+++ b/prow/config/templates/ing-Ingress.yaml
@@ -24,13 +24,6 @@ spec:
   - host: {{ .Values.prow.domain }}
     http:
       paths:
-      - path: /*
-        pathType: ImplementationSpecific
-        backend:
-          service:
-            name: deck
-            port:
-              number: 80
       - path: /hook
         pathType: ImplementationSpecific
         backend:
@@ -43,6 +36,13 @@ spec:
         backend:
           service:
             name: pushgateway-external
+            port:
+              number: 80
+      - path: /*
+        pathType: ImplementationSpecific
+        backend:
+          service:
+            name: deck
             port:
               number: 80
 {{ end }}
--- a/prow/config/values.yaml
+++ b/prow/config/values.yaml
@@ -42,7 +42,7 @@ hook:
   scrapeMetrics: false
   image: gcr.io/k8s-prow/hook:v20221121-dfec9ce18d
   service:
-    type: 'LoadBalancer'
+    type: 'ClusterIP'
   serviceAccount:
     create: false
     name: ''
--- a/prow/config/Chart.yaml
+++ b/prow/config/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v2
 name: prow-config
 description: Configuration for the ACK Prow cluster
 type: application
-version: 0.3.0
+version: 0.3.1
 appVersion: "1.16.0"


prow/config/templates/ing-Ingress.yaml
```diff
@@ -45,4 +38,11 @@ spec:
             name: pushgateway-external
             port:
               number: 80
+      - path: /*
+        pathType: ImplementationSpecific
+        backend:
+          service:
+            name: deck
+            port:
+              number: 80
```
A-Hilaly, line 47: Order matters? i guess you're configuration all `/*` **except** `/metrics` and` /hooks` ?

prow/config/values.yaml
```diff
@@ -42,7 +42,7 @@ hook:
   scrapeMetrics: false
   image: gcr.io/k8s-prow/hook:v20221121-dfec9ce18d
   service:
-    type: 'LoadBalancer'
+    type: 'ClusterIP'
```
A-Hilaly, line 45: Does the ingress automatically create a LoadBalancer to manage the `http.paths`?

prow/config/templates/ing-Ingress.yaml
```diff
@@ -24,13 +24,6 @@ spec:
...
     http:
       paths:
-      - path: /*
-        pathType: ImplementationSpecific
-        backend:
-          service:
-            name: deck
-            port:
-              number: 80
       - path: /hook
```
jaypipes, line 27: Mind putting a comment in here please that states the order of these paths is important? I'd hate to have someone revisit this in a year and break things not knowing that the order was significant...

prow/config/values.yaml
jaypipes: @RedbackThomson if we're changing this from an external load balancer to cluster-internal service management, does that mean we can get rid of the [aws-load-balancer-controller dependency](https://github.com/aws-controllers-k8s/test-infra/blob/40700938cb4cfa1dfb2fae81131349b0bad5a522/infra/lib/ci-cluster.ts#L62)?

RedbackThomson: We are still using the ALB controller to create an ALB for the ingress (https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/guide/ingress/annotations/), which filters by path down to each of the services.

prow/config/templates/ing-Ingress.yaml
RedbackThomson: I think so. I think it evaluates the path in the order they're specified. Weirdly the upstream prow still has `/*` so I don't really understand how it's working for them.

jljaco: Actually I think upstream prow has `/`, not `/*`:

https://github.com/kubernetes/test-infra/blob/cfaeafa474e934d3ed0454e1bb08edb55fcebb62/config/prow/cluster/starter/starter-s3.yaml#L491

RedbackThomson: Oh interesting, I took it from their Prow canary test setup - https://github.com/kubernetes/test-infra/blob/cfaeafa474e934d3ed0454e1bb08edb55fcebb62/config/prow/cluster/ing_ingress.yaml

RedbackThomson: Going with the spec as defined by the canaries, since these have been validated as working already.

jljaco: We discussed this offline and agreed that we don't actually know which of these is best practice.  

We're going w/ the canary test setup as that one is tested.

jljaco (approved):
/lgtm

PR #305, opened 2023-02-10
Default make target to `gen-all`

A-Hilaly:
This patch defaults the Makefile target to `gen-all` which generates
prow jobs and labels configurations

 This way we will "probably" not forget to regenerate the labels
files :)

This patch also re-generated `jobs.yaml` and `labels.yaml`.


jljaco (approved):
/lgtm

PR #308, opened 2023-02-13
Prow: Add Karpenter configuration

jljaco:
Issue #, if available:

Description of changes:
The default Karpenter provisioner in our Prow deployment was not properly configured to make use of the subnets and security groups that are spun up the CDK stack.   Also, the current CDK stack is trying to use both a managed nodegroup and Karpenter, which are not compatible.

This change:
- removes the MNG entirely
- adds some parameters to the `KarpenterAddOn` for instance types, architecture, and capacity type for EC2 resources for node
- points to the subnets and security groups that are spun up by the CDK stack

### DIFF
--- a/infra/lib/ci-cluster.ts
+++ b/infra/lib/ci-cluster.ts
@@ -44,6 +44,7 @@ export class CICluster extends Construct {
     };
 
     const karpenterAddonProps = {
+      version: "v0.24.0",
       requirements: [
           {
             key: 'node.kubernetes.io/instance-type',
--- a/infra/lib/ci-cluster.ts
+++ b/infra/lib/ci-cluster.ts
@@ -57,13 +57,12 @@ export class CICluster extends Construct {
         new ImportHostedZoneProvider(props.hostedZoneId)
       )
       .addOns(
-        new blueprints.addons.VpcCniAddOn(),
-        new blueprints.addons.KarpenterAddOn(),
-        new blueprints.addons.AwsLoadBalancerControllerAddOn(),
-        new blueprints.addons.EbsCsiDriverAddOn(),
-        new blueprints.addons.ExternalDnsAddOn({
-          hostedZoneResources: [GlobalResources.HostedZone],
-        })
+        new blueprints.CertManagerAddOn,
+        new blueprints.AwsLoadBalancerControllerAddOn,
+        new blueprints.VpcCniAddOn,
+        new blueprints.KarpenterAddOn,
+        new blueprints.EbsCsiDriverAddOn,
+        new blueprints.ExternalDnsAddOn({ hostedZoneResources: [GlobalResources.HostedZone] })
       )
       .build(this, "TestInfraCluster");
 
--- a/infra/lib/ci-cluster.ts
+++ b/infra/lib/ci-cluster.ts
@@ -7,9 +7,12 @@ import {
   ProwGitHubSecretsChartProps,
 } from "./charts/prow-secrets";
 import {
+  STACK_NAME,
   FLUX_NAMESPACE,
   PROW_JOB_NAMESPACE,
   PROW_NAMESPACE,
+  CLUSTER_NAME,
+  CLUSTER_CONSTRUCT_NAME,
 } from "./test-ci-stack";
 import {
   GlobalResources,
@@ -34,6 +37,9 @@ export class CICluster extends Construct {
 
     const clusterVersion = eks.KubernetesVersion.V1_23;
 
+    const subnetTagPattern = `${STACK_NAME}/${CLUSTER_CONSTRUCT_NAME}/${CLUSTER_NAME}/${CLUSTER_NAME}-vpc/PrivateSubnet*`;
+    const securityGroupTagPattern = "kubernetes.io/cluster/" + CLUSTER_NAME;
+
     const karpenterAddonProps = {
       requirements: [
           {
@@ -52,12 +58,8 @@ export class CICluster extends Construct {
             vals: ['on-demand']
           },
       ],
-      subnetTags: {
-        "Name": "TestCIStack/CIClusterConstruct/TestInfraCluster/TestInfraCluster-vpc/PrivateSubnet*",
-      },
-      securityGroupTags: {
-        "kubernetes.io/cluster/TestInfraCluster": "owned",
-      },
+      subnetTags: { "Name": subnetTagPattern },
+      securityGroupTags: { securityGroupTagPattern : "owned" },
       taints: [{
         key: "workload",
         value: "test",
@@ -87,7 +89,7 @@ export class CICluster extends Construct {
         new blueprints.addons.EbsCsiDriverAddOn,
         new blueprints.addons.ExternalDnsAddOn({ hostedZoneResources: [GlobalResources.HostedZone] })
       )
-      .build(this, "TestInfraCluster");
+      .build(this, CLUSTER_NAME);
 
     this.testCluster = blueprintStack.getClusterInfo().cluster;
 
--- a/infra/lib/ci-cluster.ts
+++ b/infra/lib/ci-cluster.ts
@@ -34,35 +34,58 @@ export class CICluster extends Construct {
 
     const clusterVersion = eks.KubernetesVersion.V1_23;
 
-    const mngProps: blueprints.MngClusterProviderProps = {
-      minSize: 2,
-      maxSize: 8,
-      desiredSize: 2,
-      diskSize: 150,
-      version: clusterVersion,
-      instanceTypes: [
-        ec2.InstanceType.of(ec2.InstanceClass.M5, ec2.InstanceSize.XLARGE8),
+    const karpenterAddonProps = {
+      requirements: [
+          {
+            key: 'node.kubernetes.io/instance-type',
+            op: "In" as const,
+            vals: ['m5.xlarge','m5.2xlarge','m5.4xlarge','m5.8xlarge'],
+          },
+          {
+            key: 'kubernetes.io/arch',
+            op: "In" as const,
+            vals: ['amd64']
+          },
+          {
+            key: 'karpenter.sh/capacity-type',
+            op: "In" as const,
+            vals: ['on-demand']
+          },
       ],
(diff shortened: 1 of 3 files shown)


jljaco:
I tested this in a separate "staging" account (all prow jobs disabled).  Upon a from-scratch CDK deployment, Karpenter did come up successfully and provisioned a couple of nodes.

infra/lib/ci-cluster.ts
```diff
@@ -34,36 +34,58 @@ export class CICluster extends Construct {
...
+      subnetTags: {
+        "Name": "TestCIStack/CIClusterConstruct/TestInfraCluster/TestInfraCluster-vpc/PrivateSubnet*",
+      },
+      securityGroupTags: {
+        "kubernetes.io/cluster/TestInfraCluster": "owned",
+      },
+      taints: [{
+        key: "workload",
+        value: "test",
+        effect: "NoSchedule" as const,
```
A-Hilaly, line 64: I don't see matching tolerations on our `ProwJob` pods, are we sure we'd able to schedule pods on these nodes?

infra/lib/ci-cluster.ts
```diff
@@ -34,36 +34,58 @@ export class CICluster extends Construct {
...
+      }],
+      amiFamily: "AL2" as const,
+      consolidation: { enabled: true },
+      ttlSecondsUntilExpired: 2592000,
```
A-Hilaly, line 68: Can we write this as:
```suggestion
      ttlSecondsUntilExpired: 30 * 24 * 60 * 60, // 30 days in seconds
```

infra/lib/ci-cluster.ts
```diff
@@ -34,36 +34,58 @@ export class CICluster extends Construct {
...
       ],
-      amiType: eks.NodegroupAmiType.AL2_X86_64,
-      nodeGroupCapacityType: eks.CapacityType.ON_DEMAND,
-    };
+      subnetTags: {
+        "Name": "TestCIStack/CIClusterConstruct/TestInfraCluster/TestInfraCluster-vpc/PrivateSubnet*",
+      },
+      securityGroupTags: {
+        "kubernetes.io/cluster/TestInfraCluster": "owned",
+      },
```
A-Hilaly, line 60: Not sure if this would work, but maybe we can format those tags somehow.. Thinking maybe creating a constant "ClusterName" and reuse it also here https://github.com/aws-controllers-k8s/test-infra/blob/aba46a9873a01b9b33a67260fe56461b151d8c25/infra/lib/ci-cluster.ts#L68

A-Hilaly (commented):
Nice, this will be super helpful, especially next time we make a release :)
Left few comments inline

A-Hilaly:
>couple of nodes.

How many nodes?

infra/lib/ci-cluster.ts
jljaco: Done

Parametrized the stack name, cluster name, and cluster construct name

infra/lib/ci-cluster.ts
jljaco: Done

infra/lib/ci-cluster.ts
```diff
@@ -34,38 +37,59 @@ export class CICluster extends Construct {
...
       ],
-      amiType: eks.NodegroupAmiType.AL2_X86_64,
-      nodeGroupCapacityType: eks.CapacityType.ON_DEMAND,
-    };
+      subnetTags: { "Name": subnetTagPattern },
+      securityGroupTags: { securityGroupTagPattern : "owned" },
+      taints: [{
+        key: "workload",
+        value: "test",
+        effect: "NoSchedule" as const,
```
RedbackThomson: I don't believe we need this taint. This would be necessary if we wanted certain workloads to *not run* on the Karpenter nodes. In our case we don't discriminate between pods

infra/lib/ci-cluster.ts
jljaco: Removed the taint; we can revisit using taints + tolerations in a later iteration

jljaco:
> > couple of nodes.
> 
> How many nodes?

2

infra/lib/ci-cluster.ts
```diff
@@ -34,38 +37,59 @@ export class CICluster extends Construct {
...
+      }],
+      amiFamily: "AL2" as const,
+      consolidation: { enabled: true },
+      ttlSecondsUntilExpired: 2592000,
+      weight: 20,
+      interruptionHandling: true,
+    }
```
RedbackThomson, line 68: Could we also add total limits so that the cluster doesn't exceed too many resources. See https://karpenter.sh/preview/concepts/provisioners/#speclimitsresources

I'm not sure what an appropriate value would be, but this could be updated later if we are restricted?

infra/lib/ci-cluster.ts
```diff
@@ -34,38 +37,59 @@ export class CICluster extends Construct {
...
+      securityGroupTags: { securityGroupTagPattern : "owned" },
+      taints: [{
+        key: "workload",
+        value: "test",
+        effect: "NoSchedule" as const,
+      }],
+      amiFamily: "AL2" as const,
+      consolidation: { enabled: true },
+      ttlSecondsUntilExpired: 2592000,
+      weight: 20,
```
RedbackThomson, line 66: We shouldn't need a weight if we only have one provisioner

infra/lib/ci-cluster.ts
RedbackThomson: This value seems really really high. This would mean that if the cluster scales up with another node, just to run one more ProwJob pod, then it will wait 30 days until it finally scales it back down?

infra/lib/ci-cluster.ts
jljaco: Taint removed

infra/lib/ci-cluster.ts
jljaco: I do not see any support for spec.limits in the cdk-eks-blueprints code for the KarpenterAddOn:  https://github.com/aws-quickstart/cdk-eks-blueprints/blob/01b2bd9f488ebc2bd0e159c122591437745af178/lib/addons/karpenter/index.ts#L262
So I suspect we're going to have to just add those tweaks post-hoc by `kubectl edit...`.

infra/lib/ci-cluster.ts
```diff
@@ -34,38 +37,54 @@ export class CICluster extends Construct {
...
 
-    const mngProps: blueprints.MngClusterProviderProps = {
-      minSize: 2,
-      maxSize: 8,
-      desiredSize: 2,
-      diskSize: 150,
-      version: clusterVersion,
-      instanceTypes: [
-        ec2.InstanceType.of(ec2.InstanceClass.M5, ec2.InstanceSize.XLARGE8),
+    const subnetTagPattern = `${STACK_NAME}/${CLUSTER_CONSTRUCT_NAME}/${CLUSTER_NAME}/${CLUSTER_NAME}-vpc/PrivateSubnet*`;
```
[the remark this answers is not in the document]
jljaco: `blueprintStack` is not actually declared & defined until *after* this value is being used.  It is used in the creation of the stack.  So it's a catch 22.

infra/lib/ci-cluster.ts
jljaco: Removed

jljaco:
> This value seems really really high. This would mean that if the cluster scales up with another node, just to run one more ProwJob pod, then it will wait 30 days until it finally scales it back down?

Tuned this down to 1 hour

infra/lib/ci-cluster.ts
jljaco: Tuned this down to 1 hour

infra/lib/ci-cluster.ts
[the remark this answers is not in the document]
jljaco: I checked on the names of the subnets deployed during testing; they all looked like

`TestCIStack/CIClusterConstruct/TestInfraClusterStaging/TestInfraClusterStaging-vpc/PublicSubnet3`

That looks a lot like `$STACK_NAME/$CLUSTER_CONSTRUCT_NAME/$CLUSTER_NAME/${CLUSTER_NAME}-vpc/PublicSubnet*` to me

infra/lib/ci-cluster.ts
```diff
@@ -34,38 +37,56 @@ export class CICluster extends Construct {
...
-      ],
-      amiType: eks.NodegroupAmiType.AL2_X86_64,
-      nodeGroupCapacityType: eks.CapacityType.ON_DEMAND,
+    const subnetTagPattern = `${STACK_NAME}/${CLUSTER_CONSTRUCT_NAME}/${CLUSTER_NAME}/${CLUSTER_NAME}-vpc/PrivateSubnet*`;
+    const securityGroupTagPattern = "kubernetes.io/cluster/" + CLUSTER_NAME;
+    const securityGroupTags = {
+      [securityGroupTagPattern] : "owned",
     };
 
+    const karpenterAddonProps = {
```
A-Hilaly, line 46: Is it possible to set the karpenter version?

jljaco: Done

A-Hilaly:
I don't want to block this PR, but i sent this https://github.com/aws-quickstart/cdk-eks-blueprints/pull/600 to add limits supports for Karpenter in `cdk-eks-blueprints`. Feel free to wait/include it or add it in a future PR @jljaco

A-Hilaly:
@jljaco https://github.com/aws-quickstart/cdk-eks-blueprints/pull/600 is merged. Can we import it?

jljaco:
> @jljaco [aws-quickstart/cdk-eks-blueprints#600](https://github.com/aws-quickstart/cdk-eks-blueprints/pull/600) is merged. Can we import it?

We use released versions of `cdk-eks-blueprints`, so we'll need to wait for another release which includes that change.  Prefer to do that in a follow-up PR in the interests of getting the benefits of this more quickly (i.e., having a functioning Karpenter in our Prow)

A-Hilaly (approved):
Nice, merci beaucoup!
/lgtm

PR #313, opened 2023-02-17
Move Karpenter config to GitOps

RedbackThomson:
Description of changes:
Moves the configuration of the Karpenter provider and provisioner to a GitOps `Kustomization` - so that we aren't restrained by what the EKS blueprints project supports. It also allows us to update the configuration by making changes in Git, without requiring slow CDK deployments.

### DIFF
--- a/flux/karpenter-config/node-template.yaml
+++ b/flux/karpenter-config/node-template.yaml
@@ -4,9 +4,9 @@ metadata:
   name: prowjob-node-provider
 spec:
   securityGroupSelector:
-    kubernetes.io/cluster/TestInfraCluster: owned
+    ${tagKey}: ${tagValue}
   subnetSelector:
-    kubernetes.io/cluster/TestInfraCluster: owned
+    ${tagKey}: ${tagValue}
   amiFamily: AL2
   blockDeviceMappings:
     - deviceName: /dev/xvda
--- a/flux/karpenter-config/node-template.yaml
+++ b/flux/karpenter-config/node-template.yaml
@@ -1,0 +1,16 @@
+apiVersion: karpenter.k8s.aws/v1alpha1
+kind: AWSNodeTemplate
+metadata:
+  name: prowjob-node-provider
+spec:
+  securityGroupSelector:
+    kubernetes.io/cluster/TestInfraCluster: owned
+  subnetSelector:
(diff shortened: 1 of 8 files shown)


flux/karpenter-config/node-template.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: karpenter.k8s.aws/v1alpha1
+kind: AWSNodeTemplate
+metadata:
+  name: prowjob-node-provider
+spec:
+  securityGroupSelector:
+    kubernetes.io/cluster/TestInfraCluster: owned
+  subnetSelector:
+    kubernetes.io/cluster/TestInfraCluster: owned
```
jljaco, line 9: We should make this `TestInfraCluster` a parameter; right now it lives in CDK here https://github.com/aws-controllers-k8s/test-infra/blob/b54ff65022312f8b425279857f33891a8b542eba/infra/lib/test-ci-stack.ts#L12

How can we make it so that this doesn't need to be manually changed when we change the name of the cluster?

jljaco: Disregard this, I just noticed that you're updating it below

infra/lib/ci-cluster.ts
```diff
@@ -37,39 +35,8 @@ export class CICluster extends Construct {
 
     const clusterVersion = eks.KubernetesVersion.V1_23;
 
-    const subnetTagPattern = `${STACK_NAME}/${CLUSTER_CONSTRUCT_NAME}/${CLUSTER_NAME}/${CLUSTER_NAME}-vpc/PrivateSubnet*`;
```
jljaco, line 40: It appears that we no longer need `CLUSTER_CONSTRUCT_NAME` as well

flux/karpenter-config/node-template.yaml
RedbackThomson: I have updated the CDK to now create a new ConfigMap in the `flux-system` namespace which contains `tagKey` and `tagValue` data values. Then, when we apply the `Kustomization` which creates the `karpenter-config` resources, we use a `substituteFrom` which takes all values from that configmap and applies it into the `AWSNodeTemplate` dynamically.

jljaco (approved):
/lgtm

PR #315, opened 2023-02-20
Add Karpenter do-not-evict annotations to all jobs

RedbackThomson:
Description of changes:
The [documentation for Karpenter deprovisioning](https://karpenter.sh/preview/concepts/deprovisioning/#pod-eviction) says that we can place a `karpenter.sh/do-not-evict: "true"` annotation on any pod that we do not wish to forcefully evict in the case of node consolidation. Since we want all ProwJobs to continue until completion, this pull request annotates every job using the ProwJob `annotations` list - which applies annotations to the `ProwJob` CRD and the underlying `Pod` it creates.


jljaco (approved):
/lgtm

PR #325, opened 2023-03-07
Consolidate service postsubmit jobs

RedbackThomson:
Description of changes:
The current list of jobs has a separate section for the `ack-chart` releases (per-service controller) postsubmit jobs. This pull request consolidates those jobs underneath the existing service controller postsubmit job headers. It appears that Prow will use whatever the last definition for any given repository is, within postsubmit. This meant the `ack-chart` release jobs were being overridden by the ones below it in the `jobs.yaml` and therefore never executing.


A-Hilaly (approved):
/lgtm

PR #331, opened 2023-03-16
allow var substitution in SNS/SQS bootstraps

jaypipes:
When building e2e tests for the SNS Subscription resource, I need to create an SQS Queue that refers to an SNS Topic and in order to do that, I need to add a policy document to the Queue that references the Topic. To make that easier, this patch adds support for variable substitution for the Topic and Queue bootstrap resources in the same way that the S3 Bucket bootstrap resource allows variable substitution.


A-Hilaly (approved):
Thank you!
/lgtm

PR #332, opened 2023-03-17
set name field on __post_init__ for bootstraps

jaypipes:
Sometimes we want to refer to a previously-created Bootstrappable resource from another. This happens, for example, when setting up a Queue policy to be able to read from a particular Topic.

However, for many resources, the `name` field is set at the beginning of the `bootstrap` method, which means it isn't available to reference before bootstrapping. This patch changes that so that the `name` field for IAM Roles, SNS Topics, S3 Buckets, SNS Topics and SQS Queues gets set with the `@dataclass.__post_init__` special method. This makes these field values available for referencing resources before the bootstrap phase.

Issue aws-controllers-k8s/community#1491


A-Hilaly (approved):
👍

PR #334, opened 2023-03-22
Remove `wrapper.sh` from `test-metadata-file` prow jobs

A-Hilaly:
Looks like the `wrapper.sh` is used to spin a dind, and it requires
docker to be installed which broke our first attempt to validate the
`metadata.yaml` files.

This patch fixes the errors by executing the `test-metadata-file.sh`
scripts directly from `bash`


RedbackThomson (approved):
/lgtm

PR #338, opened 2023-04-13
Add kind-dev command to start a kind dev cluster

Julian-Chu:
Description of changes:
Add `make kind-dev SERVICE=$SERVICE` to setup a cluster for development or debug quickly.

the command will execute the following actions: 
- create kind cluster
- install CRD and RBAC into cluster without controller

### DIFF
--- a/scripts/run-dev.sh
+++ b/scripts/run-dev.sh
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 
-# ./scripts/run-e2e-tests.sh is the entrypoint for ACK integration testing. It
-# ensures that a K8s cluster is acsessible, then configures the ACK controller
-# under test onto the cluster before finally running the Python tests.
+# ./scripts/run-dev.sh quickly setup cluster for ACK development. It
+# ensures that a K8s cluster is accessible, then configures the ACK CRD, RBAC
+# but not install the controller
 
 set -Eeo pipefail
 
--- a/scripts/run-e2e-tests.sh
+++ b/scripts/run-e2e-tests.sh
@@ -30,16 +30,29 @@ source "$SCRIPTS_DIR/pytest-image-runner.sh"
 
 ensure_cluster() {
     local cluster_create="$(get_cluster_create)"
-    if [[ "$cluster_create" == true ]]; then
-        local cluster_name=$(_get_kind_cluster_name)
+
+    if [[ "$cluster_create" != true ]]; then
+        info_msg "Testing connection to existing cluster ..."
+        _ensure_existing_context
+    else
+        local cluster_name=$(get_cluster_name)
+        local controller_install=${2:-true}
+        # naming cluster in the following order:
+        # 1. cluster_name is set in config file
+        # 2. cluster_name is passed as function argument. using this for dev cluster
+        # 3. generate cluster_name with random prefix
+        [ -z "$cluster_name" ] && cluster_name=$1
+        [ -z "$cluster_name" ] && cluster_name=$(_get_kind_cluster_name)
 
         info_msg "Creating KIND cluster ..."
         setup_kind_cluster "$cluster_name" "$CONTROLLER_NAMESPACE"
 
-        build_and_install_controller "$cluster_name" "$CONTROLLER_NAMESPACE" "$CONTROLLER_IMAGE_TAG"
-    else
-        info_msg "Testing connection to existing cluster ..."
-        _ensure_existing_context
+        info_msg "Installing CRDs , common and RBAC manifest..."
+        install_crd_and_rbac "$CONTROLLER_NAMESPACE"
+
+        if [[ "$controller_install" == true ]]; then
+            build_and_install_controller "$cluster_name" "$CONTROLLER_NAMESPACE" "$CONTROLLER_IMAGE_TAG"
+        fi
     fi
 }
 
@@ -92,7 +105,7 @@ _ensure_existing_context() {
 
 run() {
     ensure_aws_credentials
-    ensure_cluster
+    ensure_cluster ""
     
     build_and_run_tests
     exit $?
--- a/scripts/run-e2e-tests.sh
+++ b/scripts/run-e2e-tests.sh
@@ -29,19 +29,20 @@ source "$SCRIPTS_DIR/kind.sh"
 source "$SCRIPTS_DIR/pytest-image-runner.sh"
 
 ensure_cluster() {
+    local cluster_name_via_arg=${1:-""}
+    local controller_install=${2:-true}
     local cluster_create="$(get_cluster_create)"
+    local cluster_name=$(get_cluster_name)
 
     if [[ "$cluster_create" != true ]]; then
         info_msg "Testing connection to existing cluster ..."
         _ensure_existing_context
     else
-        local cluster_name=$(get_cluster_name)
-        local controller_install=${2:-true}
         # naming cluster in the following order:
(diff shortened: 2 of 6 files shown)


scripts/run-dev.sh
RedbackThomson, line 1: I don't think `run-dev` is a very descriptive name for this file. Could you please rename it to something like `setup-kind-cluster` or `create-dev-cluster`?

scripts/run-dev.sh
```diff
@@ -0,0 +1,63 @@
...
+source "$SCRIPTS_DIR/controller-setup.sh"
+source "$SCRIPTS_DIR/kind.sh"
+
+ensure_cluster() {
+      info_msg "Creating KIND cluster ..."
+      setup_kind_cluster "$CLUSTER_NAME" "$CONTROLLER_NAMESPACE"
+
+      info_msg "Installing CRDs , common and RBAC manifest..."
+      install_crd_and_rbac "$CONTROLLER_NAMESPACE"
+}
```
RedbackThomson, line 32: We already have an `ensure_cluster` method in another file that achieves a similar premise - https://github.com/aws-controllers-k8s/test-infra/blob/main/scripts/run-e2e-tests.sh#L31-L44

This method also uses the config YAML to determine the name, in case the user wants to use a statically named cluster. Rather than redefining this method here, maybe you could modify that one to have a similar effect and then import and use it in this script.

scripts/run-dev.sh
```diff
@@ -0,0 +1,63 @@
...
+
+run() {
+    ensure_aws_credentials
+    ensure_cluster
+    local kubeconfig_path="$ROOT_DIR/build/clusters/$CLUSTER_NAME/kubeconfig"
+    info_msg "Before running the controller, you need kubeconfig and aws credentials."
```
RedbackThomson, line 38: nit: Probably don't need this line of explanation. I think we can assume anyone using this method already knows this

scripts/run-dev.sh
```diff
@@ -0,0 +1,63 @@
...
+    info_msg "After executing the above source command to set the kubeconfig to environment, "
+    info_msg "run the following command to start the controller from the controller repo:"
+    echo ""
+    echo "go run ./cmd/controller/main.go --aws-region eu-central-1 --log-level debug --enable-development-logging"
+    echo ""
+    info_msg "if you run the controller from your code editor/IDE, you can set the following environment variables:"
+    echo "KUBECONFIG=$kubeconfig_path"
```
RedbackThomson, line 45: This is already called out from inside the `setup_kind_cluster` method. See https://github.com/aws-controllers-k8s/test-infra/blob/main/scripts/kind.sh#L27

scripts/run-dev.sh
```diff
@@ -0,0 +1,63 @@
...
+
+run() {
+    ensure_aws_credentials
+    ensure_cluster
+    local kubeconfig_path="$ROOT_DIR/build/clusters/$CLUSTER_NAME/kubeconfig"
+    info_msg "Before running the controller, you need kubeconfig and aws credentials."
+    info_msg "After executing the above source command to set the kubeconfig to environment, "
+    info_msg "run the following command to start the controller from the controller repo:"
+    echo ""
+    echo "go run ./cmd/controller/main.go --aws-region eu-central-1 --log-level debug --enable-development-logging"
```
RedbackThomson, line 42: nit: Instead of hardcoding `eu-central-1`, get the region from the config YAML ([like this](https://github.com/aws-controllers-k8s/test-infra/blob/main/scripts/run-e2e-tests.sh#L58))

RedbackThomson (changes requested):

scripts/run-dev.sh
Julian-Chu: refactor here [refactor ensure_cluster](https://github.com/aws-controllers-k8s/test-infra/pull/338/commits/7779b282dfe056b97e0c4549dddf60a6166a1958)

scripts/run-dev.sh
Julian-Chu: ya , renamed to `create-dev-cluster`

scripts/run-dev.sh
Julian-Chu: removed

scripts/run-dev.sh
Julian-Chu: I think the source command set the kubeconfig in current terminal only,  if starting the controller from goland or vscode, we have to set KUBECONFIG into them. but you're right,  better to mention in document rather than here.

scripts/run-dev.sh
Julian-Chu: fixed [rename to create-dev-cluster.sh](https://github.com/aws-controllers-k8s/test-infra/pull/338/commits/177a044a9a14e15081b3b637eada300916cced39)

A-Hilaly (approved):
Nice, good stuff!
nit: can you please update the title and description of PR to reflect your latest changes? They will be part of your git commit

Julian-Chu:
@A-Hilaly   thanks for review,  just see your comments.  the title and description have been updated!

scripts/run-e2e-tests.sh
```diff
@@ -30,16 +30,29 @@ source "$SCRIPTS_DIR/pytest-image-runner.sh"
...
     local cluster_create="$(get_cluster_create)"
-    if [[ "$cluster_create" == true ]]; then
-        local cluster_name=$(_get_kind_cluster_name)
+
+    if [[ "$cluster_create" != true ]]; then
+        info_msg "Testing connection to existing cluster ..."
+        _ensure_existing_context
+    else
+        local cluster_name=$(get_cluster_name)
+        local controller_install=${2:-true}
```
RedbackThomson, line 39: Could you please move all of the positional arguments into local variables at the top of this method, so that it's clear what the inputs are. Then you can copy them into their respective variables further down in the method as needed.

scripts/run-e2e-tests.sh
```diff
@@ -30,16 +30,29 @@ source "$SCRIPTS_DIR/pytest-image-runner.sh"
...
+        [ -z "$cluster_name" ] && cluster_name=$(_get_kind_cluster_name)
 
         info_msg "Creating KIND cluster ..."
         setup_kind_cluster "$cluster_name" "$CONTROLLER_NAMESPACE"
 
-        build_and_install_controller "$cluster_name" "$CONTROLLER_NAMESPACE" "$CONTROLLER_IMAGE_TAG"
-    else
-        info_msg "Testing connection to existing cluster ..."
-        _ensure_existing_context
+        info_msg "Installing CRDs , common and RBAC manifest..."
```
RedbackThomson, line 50:
```suggestion
info_msg "Installing CRD and RBAC manifests..."
```

Julian-Chu: fixed [4ed9f07](https://github.com/aws-controllers-k8s/test-infra/pull/338/commits/4ed9f07797baa1045da81faae0bd02f10f29738f)

scripts/run-e2e-tests.sh
Julian-Chu: moved [4c321c2](https://github.com/aws-controllers-k8s/test-infra/pull/338/commits/4c321c266b567a611a20ab7fa07da60b49e85785)

a-hilaly:
@Julian-Chu Are we ready to ship this?

a-hilaly:
Just want to double check that this is not gonna break the existing testing rule

PR #343, opened 2023-05-15
Support charts without `v` prefix in `ack-chart` CD

RedbackThomson:
Description of changes:
Removes the substitution used to remove the `v` when comparing and updating the `ack-chart` dependencies as part of its CD scripts.


a-hilaly (approved):
+++

PR #346, opened 2023-05-16
Revert "Support charts without `v` prefix in `ack-chart` CD"

RedbackThomson:
Reverts aws-controllers-k8s/test-infra#343

If the charts don't have a `v` prefix, then this code does nothing. However, if the previous charts do have a `v` prefix then this code strips it correctly. There is no reason to remove this, and it is actually necessary for migration away from `v` prefix.


jljaco (approved):
/lgtm

PR #355, opened 2023-05-18
Remove codespell Prow Job

RedbackThomson:
Description of changes:
Until we fine-tune the codespell configuration in every controller repository, so that we don't get false positives (like fIter -> filter), we should not run the codespell jobs as part of every pull request. It's nice to have spelling checks, but it should not be breaking development.


jljaco (approved):

A-Hilaly (approved):
/lgtm

PR #362, opened 2023-06-16
Temporarily pin release Go version to 1.19.8

RedbackThomson:
Description of changes:
There appears to be an error in the latest version of Golang (1.19.10) that causes `go get` and `go mod download` to pause indefinitely, without any logs, in the release postsubmit jobs. This is causing >20 releases to get stuck.

This PR modifies the GOLANG_VERSION environment variable in those jobs, hardcoding it to a known good 1.19.8, so that they can be released properly. This change should be reverted once we have validated that 1.19.11 is released and working properly.


A-Hilaly (approved):
👍 👍 
/lgtm

PR #368, opened 2023-07-18
Upgrade ack-test `PyYAML` dependency

a-hilaly:
Fixes https://github.com/aws-controllers-k8s/community/issues/1848

For the last two days, our prow jobs were failing due to some package
installation errors. This was mainly due to deprecations in the pip3 CLI
tool. The deprecated features could no longer install PyYAML v5.0.X.

This patch upgrades `PyYAML` version to `6.0.2`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 boto3==1.26.54
 kubernetes==12.0.1
-PyYAML==5.4
+PyYAML==6.0.1
 pytest-xdist==2.2.0


requirements.txt
```diff
@@ -1,4 +1,4 @@
-boto3==1.26.54
-kubernetes==12.0.1
-PyYAML==5.4
-pytest-xdist==2.2.0
+boto3==1.28.4
+kubernetes==27.2.0
```
jljaco, line 2: As discussed offline, this is a big version bump and we should do a little bit of validation to make sure that we're not going to break everything with this upgrade.  Run some e2e tests locally to lend confidence to this, document in PR description, then I think we're g2g

RedbackThomson:
I feel more confident with just updating one dependency at a time.

/lgtm

requirements.txt
a-hilaly: Reduce  the change to only update `PyYAML` to reduce risk. Didn't notice anything break with this upgrade.

PR #384, opened 2023-09-14
Add new `area` and `need-X` labels

a-hilaly:
Issue #, if available:

Description of changes:

Add these new labels:
- `area/helm`: Issues or PRs related to Helm charts
- `area/testing`: Issues or PRs related to testing
- `area/resource-references`: Issues or PRs related to resource references
- `need-investigation`: Indicates an issue needs some investigation.


prow/jobs/jinja/config/label_sync.jinja2
```diff
@@ -322,6 +337,11 @@ default:
       target: prs
       prowPlugin: require-matching-label
       addedBy: prow
+    - color: e11d21
+      description: Indicates an issue needs some investigation.
+      name: needs-investigation
```
jljaco, line 342: `priority/needs-investigation`?

prow/jobs/jinja/config/label_sync.jinja2
```diff
@@ -114,6 +124,11 @@ default:
       name: area/security
       target: both
       addedBy: label
+    - color: 0052cc
+      description: Issues or PRs related to testing
+      name: area/testing
```
jljaco, line 129: seems pretty close to `area/infra` in actual usage.

jljaco (approved):
Approved with some minor nits

prow/jobs/jinja/config/label_sync.jinja2
a-hilaly: We have few issues that are related to testing without necessarily being related to prow or infrastructure. `area/infra` is a very generalistic one that don't necessarily include testing. I personally think label will be useful.

prow/jobs/jinja/config/label_sync.jinja2
a-hilaly: We have the same `priority/*` labels as the ones defined in kubernetes. I think we should keep it that way + we already have `priority/awaiting-more-evidence` which doesn't necessarily mean that we *need* to investigate an issue. Thoughts?

PR #392, opened 2023-11-15
Bump Go compiler used to release controllers binaries to 1.21.5

a-hilaly:
We're currently using Go 1.21.0 to build ACK controllers binaries which
are shipped within ACK official official images. This bump protects us
from exposing vulnerable applications (and getting flagged by CVE scanners)
 by keeping up-to-date with the latest available Go version.

A better solution is being built that will automatically bump Go
versions in test-infra repository.


acornett21 (approved):

acornett21:
@a-hilaly Do we need to be pinned to minor versions? Can we be pinned to say `1.21` then on rebuilds we automatically get the fixes in the `dot` version? (non-blocking question)

a-hilaly:
>@a-hilaly Do we need to be pinned to minor versions? Can we be pinned to say 1.21 then on rebuilds we automatically get the fixes in the dot version? (non-blocking question)

I wish googlestorage supported that, but [`1.21` does work here](https://github.com/aws-controllers-k8s/test-infra/blob/main/prow/jobs/images/Dockerfile.deploy#L31) - we can change the way we're doing things in the future.

acornett21 (approved):

ivelichkovich (approved):

PR #327, opened 2023-03-09
feat(elbv2): elbv2 bootstrap added

ib-ak:
Issue #, if available:
Provides option to create Load Balancer for VPC Endpoint Service validation. 

Description of changes:

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #407, opened 2024-01-03
fix(elbv2): remove nlb dependency on get_bootstrap_resources()

marcdavoli:
Description of changes:
#327 added the ELBv2 / Network Load Balancer bootstrap resource, which is useful for testing resources like the VPC Endpoint Service. This PR improves upon it by removing the dependency of that bootstrap resource on the "Shared Test VPC" and the `get_bootstrap_resources()` method. In doing so, we avoid circular dependency errors in the `$SERVICE-controller` repo's test frameworks that will use the NLB bootstrap resource.


a-hilaly (approved):
Thank you!

PR #409, opened 2024-01-09
Tweak integration tests CPU/memory requests and limits

a-hilaly:
This patch fixes an inconsistency issue inthe execution time of our e2e
tests across prow jobs. Previously we noticed variations in the number
of threads used to execute the tests due to different CPU and memory
requests/limits (1 and 8) in our prow job configurations. These
variations were caused by Karpenter's behavior in selecting node sizes
based on these values, resulting in different CPU core and numbers...
and consequently, different thread counts.

This change sets more reasonable values to the memory requests and
limits for the e2e prow jobs.


jonathan-innis (approved):
/lgtm

PR #448, opened 2024-03-25
Optimize doc-releases: Schedule once per day instead of per release

a-hilaly:
To avoiid exhausting GitHub API quota when releasing multiple controllers
in parallel, the docsrelease script will now run once every 24 hours instead
of on "every controller release". With 42 controllers, the previous approach
was leading to excessive GitHub API requests due to the script crawling
and high API usage. This change optimizes the process while maintaining
regular doc updates.


prow/jobs/jinja/periodics/docs_release.jinja2
```diff
@@ -0,0 +1,32 @@
+name: docs-release
+  decorate: true
+  interval: 6h
```
acornett21, line 3: Shouldn't this be 24h?

acornett21 (commented):
Along with the question on one of the files, will this help to increase the threshold of other jobs/processes that were also experiencing rate-limiting?

prow/jobs/jinja/periodics/docs_release.jinja2
a-hilaly: Good catch! thank you sir! :)

a-hilaly:
>will this help to increase the threshold of other jobs/processes that were also experiencing rate-limiting?

@acornett21 correct, this will leave some air for other prow jobs (e.g. OLM scripts) to execute without hitting API rate limits

a-hilaly:
Note to reviewers: We will still be able to execute this job on-demand, if we don't want to wait 24hours...

prow/jobs/jinja/periodics/docs_release.jinja2
acornett21: I believe there is 24h, in another place as well.

a-hilaly: Yes we have other jobs that are scheduled to run every 6hours (label_sync, github issues stale/rotten plugins)

acornett21: Thanks for explaining, just wanted to confirm this is now 💯

PR #449, opened 2024-03-28
Fix periodic job formatting in `docs_release.jinja2`

a-hilaly:
Our last change added the `docs-release` job as a key/value map, causing a
yaml syntax issue when Prow tried to parse jobs.yaml. This commit reformats
the docs-release job in docs_release.jinja2 to be an array element instead.


acornett21:
Is there a jinja2 linter that could be added as a prehook for pr's here?

jlbutler:
+1 to linter advice, looks like there are some out there.

a-hilaly:
@acornett21 Good call! We could also just leverage `yq` to check the generate yaml directly.

acornett21:
`yq` seems like a valid option if all we care about is `is it yaml`.

jlbutler:
Let's consider using something like djLint, and maybe a preflight / test for these PRs in future.

I checked out PR and other than some old double-line warnings these look fine with `djlint` except some crufty double blank line warnings in `prow/jobs/jobs.yaml` unrelated to this PR. If you want to make it clean clean, we can remove those.

jlbutler:
We can leave off the pre-existing double-line thing for now.

/lgtm

jlbutler:
Btw I think we could also get away with `yq` - all djLint is doing here is checking yaml, as far as we're concerned. Let's just put something in here so we can catch the basic stuff.

PR #468, opened 2024-08-12
Fix ECR public session token expiration in ack-discover

a-hilaly:
These changes improve the robustness of the ack-discover tool when dealing with
longg running operations that may exceed the 1 hour session token validity

- Move ECR Public client initialization into collect_all function
- Implement re fetching of ECR public client every 58 minutes
- Remove try/except block for ECR public client initialization in main script
- Update function signature and call to `collect_all` in main script

### DIFF
--- a/tools/ackdiscover/controller.py
+++ b/tools/ackdiscover/controller.py
@@ -13,6 +13,8 @@
 
 import dataclasses
 import os
+import sys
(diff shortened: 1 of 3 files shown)


michaelhtm (approved):
/lgtm

PR #492, opened 2024-08-19
Fix kaniko dependency issues and github source files path

michaelhtm:
Description of changes:
The issue was, kaniko is not meant to be run as a binary, so we were 
having failures when building images with `debian` base image.

We are now trying to use a workaround by adding a shell to the latest 
kaniko release and running our ack-build-tools command in it to compare 
prow job image versions and update if needed.

Also fixed source files path, by making it an absolute path relative to the github directory.


a-hilaly (approved):
/lgtm

PR #508, opened 2024-08-21
Switch from kaniko to buildah for building prow images

michaelhtm:
Description of changes:

This commit refactors the process of building prow images,
moving from kaniko to buildah.
The reason for this change
is because kaniko's intended use is to build a single
image in a single container. We attempted doing a
workaround and attempt building prow images
in a single container, and the results were unpredictable.

We also considered spinning up pods/prowjobs for each
image we want to build. We decided not to go forward with
that upproach, since building images can be unpredictable,
and the prowjob that is spinning these pods is going to have
to wait for an unknown amount of time, while considering the
the different statuses a /prowjob would be in.

With these changes we are building images with buildah, and only
attempting to push when all builds are successful.

This ensures that we will have a PR with a generated `jobs.yaml`
that reflects the latest image versions in our prow ECR repository

### DIFF
--- a/prow/jobs/images/Dockerfile.build-prow-images
+++ b/prow/jobs/images/Dockerfile.build-prow-images
@@ -1,6 +1,6 @@
 ARG GO_VERSION=1.22.5
 
-FROM golang:${GO_VERSION}-alpine AS builder
+FROM public.ecr.aws/docker/library/golang:${GO_VERSION}-alpine AS builder
 
 RUN apk add --no-cache git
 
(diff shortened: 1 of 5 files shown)


prow/jobs/images/Dockerfile.build-prow-images
```diff
@@ -1,6 +1,6 @@
 ARG GO_VERSION=1.22.5
 
-FROM golang:${GO_VERSION}-alpine AS builder
+FROM public.ecr.aws/docker/library/golang:${GO_VERSION}-alpine AS builder
```
a-hilaly, line 3: 👍

prow/jobs/images/Dockerfile.build-prow-images
```diff
@@ -14,14 +14,40 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o ack-build-tools ./prow/job
 
 FROM busybox:musl AS busybox
```
a-hilaly, line 15: Are we still using busybox?

prow/jobs/tools/cmd/command/patch_prowjobs.go
```diff
@@ -94,11 +94,16 @@ func buildProwImages(cmd *cobra.Command, args []string) error {
...
-	if err = buildImagesWithKaniko(imagesConfig.ImageRepo, tagsToBuild); err != nil {
+	if err = buildImages(tagsToBuild); err != nil {
 		return err
 	}
 	log.Println("Successfully built all images")
 
+	if err = tagAndPushImages(imagesConfig.ImageRepo, tagsToBuild); err != nil {
+		return err
+	}
+	log.Printf("Error")
```
a-hilaly, line 105: Did you mean log.Println("Tagged and pushed images successfully") ?

a-hilaly (commented):
Nice! I left a few comments/questions inline

prow/jobs/images/Dockerfile.build-prow-images
michaelhtm: no, that was left from previous one

a-hilaly: Looks like you still have it on your commit https://github.com/michaelhtm/ack-test-infra/blob/fix-build-prow-images/prow/jobs/images/Dockerfile.build-prow-images#L15

prow/jobs/tools/cmd/command/patch_prowjobs.go
michaelhtm: yes

a-hilaly (approved):
/lgtm

PR #520, opened 2024-08-22
Modify `soak-test` Dockerfile

michaelhtm:
Description of changes:

I am including the dependencies that are brought from the `deploy` 
built image into `soak-test` dockerfile to allow `build-prow-images` to 
build any individual image, without it having to depend on the other images 
in `images_config.yaml`

Doing so, we can be confident that when we attempt to increase the version of a single image, we can build and push it independently, without having to rely on local base images.


a-hilaly (approved):
/lgtm

PR #524, opened 2024-08-23
Add `upgrade-eks-distro-version` prow job

michaelhtm:
Description of changes:

This prow job acts similarly to `upgrade-go-version`. With this prow job, 
we will ensure that we have the latest build images, and in the future we 
will add any other dependencies we want to keep an eye on.

This PR also contains some refactoring done to `upgrade-go-version` to allow reusability.
We are also using the `upgrade-go-version` image for this job.

In the future we will see if we can combine these two jobs into one

### DIFF
--- a/build_config.yaml
+++ b/build_config.yaml
@@ -1,1 +1,2 @@
 go_version: 1.23.0
+eks_distro_version: 2024-04-01-1711929684.2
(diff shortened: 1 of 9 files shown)


a-hilaly (approved):
Nice! Discussed some minors to be addressed in a future PR.
/lgtm

PR #533, opened 2024-08-29
Change `scan-controllers-cve` image build context

michaelhtm:
Description of changes:

Currently `prow/jobs/images` is the default prow image
build context directory. Prow images that use the `ack-build-tools`
command use the repo home directory as their build context, 
so whenever we add a command we have to change the code 
logic to include this image with the current ones that are being 
built in the `test-infra` root directory.

In the future, we want to start including the build context of the 
images we're building as part of the `images_config.yaml` file, 
as well as any other image specific information we may use.


a-hilaly (approved):
/lgtm

PR #545, opened 2024-08-29
Add build-args when building prow images

michaelhtm:
Description of changes:

We are adding `GO_VERSION` and `EKS_DISTRO_VERSION` 
build args when building our prow images. These build args will 
be used by the `deploy` image to build and release controllers 
with the updated given versions.

Later we need to make a change to `release_controller.sh` 
to allow it to pass EKS_DISTRO_VERSION as a build arg when 
building our controllers


a-hilaly (approved):
/lgtm

a-hilaly:
We're release a lot of controllers today, holding this PR just in case

PR #553, opened 2024-08-30
Change base image for `verify-attribution` job to debian

michaelhtm:
Description of changes:

Some commands in the script don't work in sh
So we're reverting back to bash and we need debian to solve this issue


a-hilaly (approved):
/lgtm

PR #584, opened 2024-11-01
Bootstrap lambda function

michaelhtm:
Description of changes:
Current usage in elbv2 controller target_group test

### DIFF
--- a/src/acktest/bootstrapping/function.py
+++ b/src/acktest/bootstrapping/function.py
@@ -1,0 +1,57 @@
+import boto3
+
+from dataclasses import dataclass, field
+
+from . import Bootstrappable
+from .. import resources
+
+@dataclass
+class Function(Bootstrappable):
+    # Inputs
+    name_prefix: str
+    code_uri: str
+    description: str = ""
+    service: str
+
+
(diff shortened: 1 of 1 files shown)


src/acktest/bootstrapping/function.py
```diff
@@ -0,0 +1,57 @@
...
+        )
+
+        self.arn = function["FunctionArn"]
+
+        self.lambda_client.add_permission(
+            FunctionName=self.name,
+            StatementId=f"{self.name}-invoke",
+            SourceArn=f"arn:aws:{self.service}:{self.region}:{resources.get_account_id()}:*",
+            Action="lambda:InvokeFunction",
+            Principal="elasticloadbalancer.amazonaws.com"
```
a-hilaly, line 48: nit: Ideally we would want this to be configurable, in case we want to use the lambda with something else.

a-hilaly (approved):
👍 
/lgtm

src/acktest/bootstrapping/function.py
a-hilaly: maybe to add in the future @michaelhtm

PR #617, opened 2025-01-23
Correct CPU resource limit unit in service tests template

michaelhtm:
Changed CPU resource limit from '16gi' to '16' to use correct unit.
CPU resources are specified in cores/millicores while 'gi' is
for memory units. This fixes the invalid resource specification.


a-hilaly (approved):
/lgtm

PR #621, opened 2025-02-05
Pass the `LOCAL_MODULES` env to `build-controller-image.sh`

itaiatu:
Currently, `LOCAL_MODULES` is not passed to the `build-controller-image.sh` and we cannot test the local `runtime` changes in order to build the controller's image with the `Dockerfile.local` file.

Putting it in front of the command, it will be passed to the `build-controller-image.sh` script and it will use the local `runtime` code.

### DIFF
--- a/scripts/controller-setup.sh
+++ b/scripts/controller-setup.sh
@@ -48,9 +48,8 @@ _build_controller_image() {
     local __img_name=$1
 
     local local_build_default="$(get_is_local_build)"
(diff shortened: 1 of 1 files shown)


JonnieDoe (approved):
/lgtm

a-hilaly (approved):
👍

michaelhtm (commented):
👍

PR #639, opened 2025-03-18
Enable Addoption feature flag on all tests

michaelhtm:
Description of changes:
It has been counter productive to have to update the list
of service controllers that have this feature enabled whenever
we need to add tests for them. This change ensures the feature
is enabled by default in all resources during tests, so if we
need to add addoption tetsts to other services, we can do so
without having to update the `AddoptionTestServices` list


a-hilaly (approved):
Thanks @michaelhtm !
/lgtm

PR #643, opened 2025-03-27
fix: increase memory limits for code generation jobs

a-hilaly:
Increases memory allocation for code generation jobs to address OOM failures
occurring after recent dependency updates.

Would be cool to see what we can do with DRA.


rushmash91 (approved):
/lgtm

PR #673, opened 2025-05-30
Apply Github CLI workaround

knottnt:
Description of changes:
- Apply Github CLI workaround described in the issue.
- Upgrade debian base images to get latest Git version with support for `--show-current`.


prow/jobs/images/Dockerfile.auto-generate-controllers
```diff
@@ -1,4 +1,4 @@
-FROM debian:buster-slim AS base
+FROM debian:bookworm-slim AS base
```
knottnt, line 1: Upgrading to bookworm as the buster image provides an outdated version of git which doesn't support the `--show-current` flag. I've upgraded the other job images as well since `debian:buster` is EOL.

PR #693, opened 2025-06-19
add resource addition workflow

rushmash91:
Description of changes:
- add a workflow to call agents and make then reusable
- remove unused tools from generator agent
- update generator to not call model agent 


experimental/agents/workflows/ack_resource_workflow.py
```diff
@@ -0,0 +1,349 @@
...
+    search_memories,
+    update_service_generator_config,
+)
+
+@dataclass
+class ResourceAdditionInput:
+    """Input for the ACK Resource Addition Workflow."""
+    service: str
+    resource: str
+    aws_sdk_version: str = "v1.32.6"
```
knottnt, line 50: Should we just let the ack-generate metadata set this? Maybe this could off by default, but added as an override?

knottnt (approved):
/lgtm

PR #700, opened 2025-07-08
Add VPC Endpoint Service as bootstrappable resource

knottnt:
Issue #, if available:

Description of changes:
- Add VpcEndpointServiceConfiguration bootstrappable dataclass

### DIFF
--- a/src/acktest/bootstrapping/vpc_endpoint_service.py
+++ b/src/acktest/bootstrapping/vpc_endpoint_service.py
@@ -1,0 +1,58 @@
+import boto3
+
+from dataclasses import dataclass, field
+from acktest.bootstrapping.elbv2 import NetworkLoadBalancer
+from . import Bootstrappable
+from .. import resources
+
+@dataclass
+class VpcEndpointServiceConfiguration(Bootstrappable):
+    # Inputs
+    name_prefix: str
+    name: str = field(init=False, default=None)
+
+    # Subresources
+    networkLoadBalancer: NetworkLoadBalancer = field(init=False, default=None)
+
+    # Outputs
+    service_id: str = field(init=False)
+
+    def __post_init__(self):
+        self.name = resources.random_suffix_name(self.name_prefix, 24)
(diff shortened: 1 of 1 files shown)


a-hilaly (approved):
👍

src/acktest/bootstrapping/vpc_endpoint_service.py
```diff
@@ -0,0 +1,58 @@
...
+                    }
+                ]
+            }]
+        )
+
+        self.service_id = vpc_endpoint_service["ServiceConfiguration"]["ServiceId"]
+
+    
+    def cleanup(self):
+        if hasattr(self, "service_id"):
```
michaelhtm, line 54: nit: note sure if this is necessary..i don't see a case where this would ever be false

src/acktest/bootstrapping/vpc_endpoint_service.py
```diff
@@ -0,0 +1,58 @@
...
+            NetworkLoadBalancerArns=[self.networkLoadBalancer.arn],
+            TagSpecifications=[{
+                "ResourceType": "vpc-endpoint-service",
+                "Tags": [
+                    {
+                        "Key": "Name",
+                        "Value": self.name
+                    }
+                ]
+            }]
```
michaelhtm, line 47: I don't see a need for defining tags here..

michaelhtm (approved):
lgtm Overall! just some minor nits

src/acktest/bootstrapping/vpc_endpoint_service.py
knottnt: It's mainly to help identify test resources. In the console the 'Name' tag is displayed in the 'name' column.

src/acktest/bootstrapping/vpc_endpoint_service.py
knottnt: The `service_id` field has `init=False` set which means that the field won't be set in the classes `__init__()` method. So, if there is a failure prior to the field being set in `bootstrap()` the field is not created.

michaelhtm (approved):
/lgtm

PR #708, opened 2025-07-31
chore: add partition to acktest identity functions

michaelhtm:

### DIFF
--- a/src/acktest/aws/identity.py
+++ b/src/acktest/aws/identity.py
@@ -22,3 +22,6 @@ def get_account_id() -> int:
 
 def get_region(default: str = "us-west-2") -> str:
     return boto3.session.Session().region_name or default
+
+def get_partition() -> str:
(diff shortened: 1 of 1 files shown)


src/acktest/aws/identity.py
```diff
@@ -22,3 +22,6 @@ def get_account_id() -> int:
 
 def get_region(default: str = "us-west-2") -> str:
     return boto3.session.Session().region_name or default
+
+def get_partition() -> str:
+    return boto3.client('sts').get_caller_identity().get('Arn').split(':')[1]
```
a-hilaly, line 27: how about using botocore.arn/parse_arn https://github.com/boto/botocore/blob/develop/botocore/utils.py#L2136

michaelhtm: I did see that..it just requires us to have a new dependency in requirements.txt just to get partition from an arn.
I am willing to add it if you think its best

PR #726, opened 2025-08-13
Set downstream controller version from PR labels for auto-generate-controller job

knottnt:
Issue #, if available:

Description of changes:
- Add labels indicating patch and minor version for downstream controller update from code-generator releases
- Update auto-generate-controllers.sh to retrieve PR labels for latest tagged commit and base controller version bump off of version/minor and version/patch labels.


cd/auto-generate/auto-generate-controllers.sh
```diff
@@ -90,6 +112,22 @@ else
...
+PR_LABELS=$(get_pr_labels)
+
+# Set type of version bump to apply to downstream controller releases.
+CONTROLLER_RELEASE_TYPE=""
+if echo "$labels" | grep -q "version/patch"; then
+    CONTROLLER_RELEASE_TYPE="patch"
+elif echo "$labels" | grep -q "version/minor"; then
+    CONTROLLER_RELEASE_TYPE="minor"
+else 
+    echo "auto-generate-controllers.sh][ERROR] No labels specifying controller release type found exiting."
```
michaelhtm, line 125: would failing postsubmit job be too aggressive? maybe we can block merging release PR instead unless there is a version defined

cd/auto-generate/auto-generate-controllers.sh
```diff
@@ -90,6 +112,22 @@ else
...
+    exit 1
+fi
+
+echo "auto-generate-controllers.sh][INFO] applying $CONTROLLER_RELEASE_TYPE version bump to downstream controller repos."
```
michaelhtm, line 129: Nice

prow/jobs/images_config.yaml
```diff
@@ -1,6 +1,6 @@
 image_repo: public.ecr.aws/m5q3e4b2/prow
 images:
-    auto-generate-controllers: prow-auto-generate-controllers-0.0.29
+    auto-generate-controllers: prow-auto-generate-controllers-0.0.30
```
michaelhtm, line 3: This job does not require being rebuilt (it does not copy the script in the image..it uses latest one)

cd/auto-generate/auto-generate-controllers.sh
michaelhtm: or we can use patch by default, unless there is the `minor` label..

michaelhtm (commented):
Great work @knottnt 🙌 
left a few comments

cd/auto-generate/auto-generate-controllers.sh
knottnt: I considered just patching by default. The reason I leaned towards failing early is that I wanted the developer making the release to actively make choose what kind of downstream version bump should be used. If someone forgets to add the label they can go back and add it after to the PR and restart the job.

prow/jobs/images_config.yaml
knottnt: Okay, will remove.

cd/auto-generate/auto-generate-controllers.sh
michaelhtm: restarting a job is not really the best experience..maybe we can add a `hold/includeDownstreamVersionBeforeMerge` label when PR is created by default that blocks merge (added here: https://github.com/aws-controllers-k8s/test-infra/blob/f6c882880bbca6c7e248923077b701879157ee44/prow/config/templates/config-ConfigMap.yaml#L97-L101).

knottnt: Maybe, how fine grained is that tooling? We really only want to have that requirement for release PRs. The labels don't have any meaning to most code-gen PRs.

michaelhtm (approved):
/lgtm

PR #900, opened 2026-04-21
fix: validate ack-generate-metadata.yaml version field during code-gen

michaelhtm:
Description of changes:
The previous filter used a Perl regex lookahead that was incompatible with
grep -E, silently skipping version validation. Replace it with an explicit
validation step that extracts the semver prefix (vX.Y.Z) from both the
committed and regenerated files and fails if they differ or either contains
a 'dirty' suffix. The version line is then filtered as non-deterministic
noise only after validation passes.

### DIFF
--- a/cd/scripts/verify-code-gen.sh
+++ b/cd/scripts/verify-code-gen.sh
@@ -69,13 +69,58 @@ info_msg "Checking for differences..."
 # non-deterministic patterns scoped to specific files. If no meaningful
 # changes remain, skip the file.
 #
+# ack-generate-metadata.yaml gets special handling: the version field is
+# validated (semver prefix must match, 'dirty' suffix is rejected) before
+# the git-describe suffix is filtered out as noise.
+#
 # Patterns are applied only to the files that contain them:
 #   ack-generate-metadata.yaml: api_directory_checksum, build_date,
-#                                build_hash, go_version
+#                                build_hash, go_version, version
 #   kustomization.yaml:         newTag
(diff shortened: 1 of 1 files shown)


cd/scripts/verify-code-gen.sh
```diff
@@ -69,22 +69,71 @@ info_msg "Checking for differences..."
...
+        error_msg "Could not read committed version from $file"
+        exit 1
+    fi
+    if [ -z "$regenerated" ]; then
+        error_msg "Could not read regenerated version from $file"
+        exit 1
+    fi
+
+    if echo "$committed" | grep -q 'dirty'; then
+        error_msg "Committed $file has a dirty version: $committed"
```
knottnt: nit: It's pretty common for new contributors to run into this. Adding something like "ensure that all changes in local code-generator repo have been committed upstream" could be helpful.

cd/scripts/verify-code-gen.sh
```diff
@@ -69,22 +69,71 @@ info_msg "Checking for differences..."
...
+        exit 1
+    fi
+    if echo "$regenerated" | grep -q 'dirty'; then
+        error_msg "Regenerated $file has a dirty version: $regenerated"
```
knottnt: When would this happen? The ProwJob should always be pointed at the HEAD.

knottnt (commented):
Thanks @michaelhtm!

knottnt (approved):
Nice!
/lgtm

PR #907, opened 2026-04-28
Add MWAA bootstrapping helpers

naclonts:
Issue #, if available: aws-controllers-k8s/community#2761

Description of changes: Adds opt-in parameters to existing `acktest.bootstrapping` primitives so [MWAA](https://github.com/aws-controllers-k8s/mwaa-controller/pull/3) e2e tests can bootstrap with the stock helpers. All defaults preserve existing behavior.

  - `iam.Role`: new `additional_service_principals: List[str]` for roles that need multiple trusted services (MWAA requires both `airflow.amazonaws.com` and `airflow-env.amazonaws.com`).
  - `s3.Bucket`:
    - new `empty_objects: List[str]` to pre-create zero-byte objects after the bucket is created (MWAA `CreateEnvironment` validates that the DAGs prefix exists).
    - `cleanup()` now purges all object versions and delete markers when `enable_versioning=True`, so teardown succeeds for versioned buckets instead of failing with `BucketNotEmpty`.
  - `vpc.SecurityGroup`: new `self_referencing_ingress: bool` that adds an idempotent self-referencing inbound rule (required for MWAA worker-to-worker traffic).

### DIFF
--- a/src/acktest/bootstrapping/iam.py
+++ b/src/acktest/bootstrapping/iam.py
@@ -63,6 +63,7 @@ class Role(Bootstrappable):
     principal_service: str
     description: str = ""
     managed_policies: List[str] = field(default_factory=lambda: [])
+    additional_service_principals: List[str] = field(default_factory=lambda: [])
 
     # Subresources
     user_policies: UserPolicies = field(default=None)
@@ -78,6 +79,18 @@ class Role(Bootstrappable):
     def iam_client(self):
         return boto3.client("iam", region_name=self.region)
 
+    def _trust_policy_services(self):
+        """Returns the list of service principals for the trust policy.
+
+        When `additional_service_principals` is empty the IAM `Service` value
+        is emitted as a single string (preserving the wire format used by
+        single-principal roles); when it is non-empty the combined list is
+        emitted so IAM accepts it as a multi-principal trust.
+        """
+        if not self.additional_service_principals:
+            return self.principal_service
+        return [self.principal_service, *self.additional_service_principals]
+
     def bootstrap(self):
         """Creates an IAM role with an auto-generated name.
         """
@@ -91,7 +104,7 @@ class Role(Bootstrappable):
                     "Statement": [
                         {
                             "Effect": "Allow",
-                            "Principal": {"Service": self.principal_service},
+                            "Principal": {"Service": self._trust_policy_services()},
                             "Action": ["sts:AssumeRole", "sts:TagSession"],
                         }
                     ],
--- a/src/acktest/bootstrapping/s3.py
+++ b/src/acktest/bootstrapping/s3.py
@@ -2,6 +2,7 @@ import boto3
 import logging
 
 from dataclasses import dataclass, field
+from typing import List
 
 from . import Bootstrappable
 from .. import resources
@@ -13,6 +14,9 @@ class Bucket(Bootstrappable):
     enable_versioning: bool = False
     policy: str = ""
     policy_vars: dict = field(default_factory=dict)
+    # Optional list of object keys to pre-create as zero-byte objects
+    # after the bucket is created.
+    empty_objects: List[str] = field(default_factory=list)
 
     # Outputs
     name: str = field(init=False)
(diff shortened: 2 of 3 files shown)


src/acktest/bootstrapping/mwaa.py
```diff
@@ -0,0 +1,262 @@
...
+from typing import List
+
+from . import Bootstrappable
+from .iam import UserPolicies
+from .s3 import Bucket
+from .. import resources
+
+
+@dataclass
+class MWAAExecutionRole(Bootstrappable):
```
knottnt: Q: If the main difference is just the number of trusted services it seems like we could make this more generic and move this to iam.py. I'm sure other services in AWS will want to use this and having this under MWAA would be confusing going forward.

src/acktest/bootstrapping/mwaa.py
```diff
@@ -0,0 +1,262 @@
...
+
+    MWAA internally uses two identities — `airflow.amazonaws.com` (the
+    control-plane service) and `airflow-env.amazonaws.com` (the data-plane
+    service that runs on customer subnets). Both must be trusted by the
+    execution role or `CreateEnvironment` fails with an `AccessDeniedException`
+    on validation.
+
+    This is a purpose-built wrapper rather than a subclass of `iam.Role`
+    because `Role.principal_service` is a single string and dataclass
+    inheritance with overridden field types is awkward.
```
knottnt: Does this work with a list of one? If it does we could change the internal logic of the existing Role to use that and provide either an alternative constructor or a new optional `additional_service_principals` field to the Role dataclass.

src/acktest/bootstrapping/mwaa.py
```diff
@@ -0,0 +1,262 @@
...
+
+@dataclass
+class MWAADAGBucket(Bootstrappable):
+    """S3 bucket for MWAA DAG storage.
+
+    MWAA's `CreateEnvironment` validation rejects a `SourceBucketArn` whose
+    DAGs prefix does not already exist. This wraps the generic `s3.Bucket`
+    and, after the bucket is created, uploads a zero-byte object at the DAGs
+    prefix to satisfy that check.
+
```
knottnt: Adding an object at a specific prefix path seems like it could be more generic than just the MWAA use case. Could this be supported as an optional parameter on the existing Bucket class in s3.py?

knottnt: Or at least renamed and moved to s3.py?

src/acktest/bootstrapping/mwaa.py
```diff
@@ -0,0 +1,262 @@
...
+                # Already gone; let Bucket.cleanup() handle the idempotent
+                # path (it currently raises, but this matches current
+                # behavior of the generic cleanup).
+                pass
+
+        self.bucket.cleanup()
+
+
+@dataclass
+class SelfReferencingSecurityGroupIngress(Bootstrappable):
```
knottnt: Could this be implemented in a way that extends VPC or SecurityGroup in vpc.py?

knottnt: Otherwise, this can could still be moved to vpc.py. The current use case is for MWAA, but this isn't a MWAA specific concept.

naclonts: Moved to an optional `self_referencing_ingress` param in vpc.py's `SecurityGroup` class.

src/acktest/bootstrapping/mwaa.py
naclonts: Moved to s3.py, as an optional `empty_objects` param which creates the requested paths as zero-byte files.

src/acktest/bootstrapping/mwaa.py
naclonts: Good call, updated the iam.py `Role` to accept an `additional_service_principals` field.

src/acktest/bootstrapping/mwaa.py
naclonts: Removed this file, adding the logic to the existing resources.

knottnt (approved):
Thanks @naclonts! 
/lgtm

PR #951, opened 2026-06-02
fix: bump all prow job image versions to trigger rebuild

gustavodiaz7722:
Bumps all prow job image versions (except `build-prow-images` which was already bumped in #948) to trigger a full rebuild of all job images from the current source code.

This ensures all images pick up the latest changes from the Flux v2 migration merge.


michaelhtm (approved):
/lgtm

PR #973, opened 2026-06-15
fix: align kubelet and containerd on cgroupfs for KIND on cgroup v1 hosts

k1ranpk:
On cgroup v1 hosts (e.g. AL2 dev desktops), `make kind-test` fails because the default KIND config
  leaves the kubelet unhealthy, and a kubelet-only cgroupfs override mismatches containerd (`runc: expected
  cgroupsPath to be of format slice:prefix:name`). This sets both kubelet and containerd to the cgroupfs driver so
  the control plane initializes.
  
  Opening for discussion — happy to make this conditional on the host cgroup version if preferred, since cgroup v2
  hosts typically default to the systemd driver.


k1ranpk:
Closing — KIND cgroupfs configuration is already present in upstream/main via #903 (Fix KIND cluster
  creation for Kubernetes v1.30+). My local clone was stale; no change needed.

PR #987, opened 2026-06-23
soak: fix parallel execution with per-service kubeconfig

gustavodiaz7722:
## Summary

Fixes a race condition when running multiple `bootstrap.sh` instances in parallel. Each instance now uses an isolated kubeconfig file, preventing kubectl context collisions.

## Problem

When launching soak tests for multiple controllers simultaneously, all `bootstrap.sh` processes call `aws eks update-kubeconfig` which writes to the shared `~/.kube/config`. This causes:
- Controllers installed in the wrong cluster
- Helm operations targeting the wrong context
- Prometheus/Loki installs failing with namespace-not-found errors

## Fix

```bash
export KUBECONFIG="${SOAK_DIR}/.kubeconfig-${AWS_SERVICE}"
```

Each bootstrap writes its kubeconfig to a service-specific file that's cleaned up on exit via the existing `trap`.

## New Scripts

- **run-all.sh**: Batch-launches soak tests for all configured controllers with optional parallelism throttling. Logs per-service output to `/tmp/soak-logs/`.
- **teardown-all.sh**: Tears down all running soak clusters in parallel.

## Testing

Launched 10 soak tests in parallel — all deployed successfully without cross-contamination:
```
autoscaling               Soak test is running
backup                    Soak test is running
dsql                      Soak test is running
emrserverless             Soak test is running
firehose                  Soak test is running
glue                      Soak test is running
mwaa                      Soak test is running
opensearchserverless      Soak test is running
quicksight                Soak test is running
s3files                   Soak test is running
```

### DIFF
--- a/soak/.gitignore
+++ b/soak/.gitignore
@@ -1,0 +1,1 @@
+.kubeconfig-*
--- a/soak/README.md
+++ b/soak/README.md
@@ -170,11 +170,41 @@ After reviewing results, clean up all resources for a specific service:
 ./teardown.sh <service>
 ```
 
+Or tear down all soak clusters at once:
+```bash
+./teardown-all.sh
+```
+
 Or manually:
 ```bash
 eksctl delete cluster --name ack-soak-<service> --region us-west-2
 aws --region us-east-1 ecr-public delete-repository --repository-name ack-<service>-soak --force
 ```
+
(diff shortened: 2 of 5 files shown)


sapphirew (approved):
/lgtm

PR #997, opened 2026-06-25
hardening: isolate Prow control plane and add PDBs for deck/hook

gustavodiaz7722:
## Description

Resilience hardening for the Prow cluster, scoped to reduce the blast radius of
node-level disruptions on the control plane.

### Background / findings

Investigation of the durable EKS control-plane logs (CloudWatch) and CloudTrail
established the following, with hard evidence:

- The control-plane pod disruptions were **voluntary node drains initiated by
  EKS Auto Mode**. Audit logs show every eviction (deck, hook, tide, crier,
  sinker, statusreconciler, horologium, prow-controller-manager, ghproxy, plus
  Prometheus/Flux/CoreDNS) was issued by
  `AWSServiceRoleForAmazonEKS/eks-auto-mode-compute`.
- The drained nodes were **healthy** when drained — the kube-controller-manager
  log shows no NotReady transition before the drain; it was an orderly
  eviction -> volume detach -> node deletion sequence.
- The trigger was **10-hour node expiry**. CloudTrail shows two drained nodes
  with `RunInstances` -> `TerminateInstances` lifetimes of exactly ~10h, matching
  the `prow-compute` NodePool's `expireAfter: 10h`:

  | Node | RunInstances | TerminateInstances | Lifetime |
  |------|-------------|--------------------|---------|
  | i-012538b7bd3697e76 | 09:47:16Z | 19:47:51Z | 10h 00m |
  | i-0fbe9a41681d0a2df | 09:50:40Z | 19:50:46Z | 10h 00m |

Because the control-plane pods were co-located on these recycling nodes with no
PodDisruptionBudget and no node isolation, each node expiry evicted control-plane
replicas together, producing brief unavailability windows.

> Note: Karpenter's internal disruption-decision log runs in the AWS-managed
> control plane and is not customer-visible, so the literal "Expired" reason
> isn't in our logs. The `expireAfter: 10h` config plus two nodes terminated at
> exactly the 10h mark by the Auto Mode role is the supporting evidence.

### Eviction / churn quantification (24h audit-log analysis)

Aggregating the kube-apiserver audit log over a 24h window:

- **52 evictions of `prow` namespace pods, 100% initiated by
  `eks-auto-mode-compute`** (node drains). Zero kubelet/resource-pressure
  evictions; zero manual evictions.
- Evictions are **batched into 4 node-drain events** (~23:00, 06:00, 09:00,
  19:00), each draining ~11–14 control-plane pods at once. deck evicted 10×,
  hook 11×.
- **Control-plane pod reschedules (~53) match the eviction count ~1:1** and
  cluster at the same 4 windows — churn is eviction-driven.
- **No rollout/config-driven churn**: each control-plane Deployment shows a
  single ReplicaSet revision across the window (no new revisions). Flux's
  periodic spec re-applies are idempotent and do not trigger restarts; the
  high count of deployment/replicaset writes is normal `status` bookkeeping.

Conclusion: control-plane churn is driven **entirely by EKS Auto Mode node
drains (~4/day)** — not resource pressure, Flux rollouts, or config changes.

### Node lifecycle: two distinct patterns

Pairing all `RunInstances`/`TerminateInstances` events by `eks-auto-mode-compute`
over an 18h window shows node lifetimes fall into two groups:

- **~10h nodes** — steady-state capacity recycled by `expireAfter: 10h` (above).
- **1–2h nodes** — short-lived burst capacity scaled up for e2e job spikes and
  released afterward. Observed lifetimes: 1.04h, 1.19h, 1.45h, 2.14h.

Traced example of the burst cycle (CloudTrail + kube-scheduler logs), node
`i-0d35f9653c1c90907`:

| Time | Event |
|------|-------|
| 23:50:39Z | RunInstances (scaled up) |
| 23:50:51Z | DaemonSets + first e2e `test-pods` bound |
| 23:50–01:01 | ~19 `test-pods` e2e jobs scheduled/run |
| 01:01:53Z | last job bound |
| 01:17:50Z | TerminateInstances (scaled down) |

The node was launched to absorb an e2e job burst and terminated ~16 min after
the last job activity, matching the NodePool's `consolidationPolicy: WhenEmpty` +
`consolidateAfter: 15m`. This scale-up/scale-down cycle is **expected, healthy
Karpenter behavior and is not a fault**. It is documented here for completeness.

Importantly, `WhenEmpty` consolidation does **not** evict nodes hosting
control-plane pods (such nodes are not "empty"), so this cycle is not an
additional control-plane disruptor — the control-plane disruptor is the 10h
expiry. The isolation below also removes any indirect coupling (a control-plane
pod transiently landing on a burst node).

### Changes

**1. Control-plane node isolation**
- New dedicated, tainted `prow-control-plane` NodePool with `expireAfter: 720h`
  (instead of the e2e pool's 10h), so the control plane is not recycled on the
  aggressive churn cadence.
- Prow control-plane deployments are pinned to it via a `nodeSelector` +
  `toleration`, applied through a shared `controlPlaneScheduling` helper.

**2. PodDisruptionBudgets for deck and hook**
- `minAvailable: 1` PDBs so a node drain (expiry, consolidation, rollout) keeps
  at least one replica serving.
- Scoped to deck and hook only — they are the only multi-replica components.
  Single-replica components (tide, ghproxy, horologium, crier, sinker,
  statusreconciler, prow-controller-manager) are intentionally excluded: a
  `minAvailable: 1` PDB on a single-replica Deployment blocks node draining
  entirely and cannot improve availability.

Chart version bumped to `0.4.19` to trigger reconciliation.

### Removed from this PR
- An earlier draft raised the compute NodeClass EBS throughput/IOPS. The logs
  showed node terminations were time-based expiry, not EBS-triggered, so that
  change is not tied to the identified behavior and has been dropped. Concurrent
  EBS throughput/IOPS-exceeded events were observed but are not implicated in the
  node churn; they can be addressed separately if warranted.

## Testing
- `kubectl kustomize flux/ack/cluster` renders cleanly.
- `helm template prow/config` renders: control-plane deployments receive the
  nodeSelector/toleration; both PDBs render.

## Notes
The `prow-control-plane` NodePool is tainted, so it stays at 0 nodes until the
pinned control-plane pods schedule onto it; the e2e pool is unaffected.


michaelhtm (approved):

knottnt (approved):
/lgtm

PR #998, opened 2026-06-26
soak: install EBS CSI driver so controller logs are persisted

gustavodiaz7722:
## Problem

The soak monitoring stack deploys Loki with a 15Gi `PersistentVolumeClaim` to persist controller and test-runner logs. But the soak EKS clusters had **no EBS CSI driver**, so the `gp2` PVC never bound — the Loki pod stayed `Pending` indefinitely and **no logs were persisted**.

The only logs available were ephemeral `kubectl logs`, which are lost when the controller pod restarts or the cluster is torn down. This makes it impossible to investigate a test failure after the fact.

Observed on running soak clusters:
```
$ kubectl get pvc -n prometheus-<service>
storage-loki-<service>-0   Pending   ...   gp2
$ kubectl get pods -n prometheus-<service> | grep loki
loki-<service>-0   0/1   Pending   0   2d21h
```
PVC events: `Waiting for a volume to be created ... by the external provisioner 'ebs.csi.aws.com'` — but the driver was never installed (EKS addons were only coredns, kube-proxy, metrics-server, vpc-cni).

## Fix

Add the `aws-ebs-csi-driver` addon to `cluster-config.yaml`, using an **EKS Pod Identity association** for its IAM permissions (`useDefaultPodIdentityAssociations: true`). The `eks-pod-identity-agent` addon is declared as the required prerequisite.

```yaml
addons:
- name: eks-pod-identity-agent
- name: aws-ebs-csi-driver
  useDefaultPodIdentityAssociations: true
```

With the CSI driver present, the Loki PVC binds, the Loki pod runs, and controller logs are persisted for the lifetime of the cluster.

Also documents in the README how to investigate test failures via Loki (Grafana Explore + Loki API), including correlating a failed test iteration with the controller's reconcile activity in the same time window.

## Testing

Verified end-to-end on a freshly bootstrapped `ack-soak-glue` cluster:
- `aws-ebs-csi-driver` addon **ACTIVE** with a pod identity association created
- `ebs-csi-controller` / `ebs-csi-node` / `eks-pod-identity-agent` pods Running
- Loki PVC **Bound** (15Gi gp2), Loki pod Running
- Controller log streams queryable from Loki, and a failed-iteration → controller-activity correlation confirmed working

### DIFF
--- a/soak/README.md
+++ b/soak/README.md
@@ -77,9 +77,11 @@ Run the `bootstrap.sh` script in the current directory. This script will do the
 following: 
 1. Create an ECR public repository for the soak test runner (`ack-<service>-soak`)
 1. Create an EKS cluster named `ack-soak-<service>` with the default soak test
-   configuration (from `cluster-config.yaml`)
+   configuration (from `cluster-config.yaml`). This includes the EBS CSI driver
+   addon (with an EKS Pod Identity association), which is required to provision
+   the persistent volume that stores controller logs in Loki.
 1. Install the controller Helm chart (`soak-test-<service>`)
-1. Install Prometheus and Grafana in namespace `prometheus-<service>`
+1. Install Prometheus, Grafana, and Loki in namespace `prometheus-<service>`
 1. Install the custom ACK soak test Grafana dashboard (from
    `./monitoring/grafana/ack-dashboard-source.json`)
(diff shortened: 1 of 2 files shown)


sapphirew (approved):
/lgtm

PR #1003, opened 2026-06-29
docs(soak): Expand soak checklist summary into evidence-log format

gustavodiaz7722:
## Description

Follow-up to #1001. Expands the soak test checklist's pass/fail summary so each conclusion is reproducible and auditable.

### Changes
- Replaces the compact pass/fail table with a **verdict table** (quick PASS/FAIL per check) plus a **per-check evidence log**. Each block captures four things:
  - **What we looked at** — the specific metric / log signal
  - **Command run** — the exact command/query used to obtain it
  - **Observed** — the value(s) returned
  - **Conclusion** — how the observed value maps to the pass criteria
- Adds a fully **worked `ack-soak-glue` example** (24h run) with real numbers, doubling as a calibration baseline for a healthy single-resource controller.

### Why
The previous table recorded only a verdict and a hint. Reviewers couldn't tell *how* a run was judged or reproduce it. The evidence-log format makes a soak sign-off self-documenting and repeatable.

### DIFF
--- a/soak/SOAK_TEST_CHECKLIST.md
+++ b/soak/SOAK_TEST_CHECKLIST.md
@@ -172,32 +172,217 @@ disproportionately (a sign of redundant Describe/List churn).
 
 ## Pass/Fail summary template
 
-| # | Check | Required | Result | Evidence |
-|---|-------|----------|--------|----------|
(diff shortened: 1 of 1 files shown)


sapphirew (approved):
/lgtm

PR #1004, opened 2026-07-01
feat(acktest): add wait_on_condition_after for fresh-reconcile waits

gustavodiaz7722:
## Description

Adds a timestamp-aware way to wait for a Kubernetes condition, so e2e tests can wait for a **fresh** reconcile after patching a resource instead of matching a condition that may still carry the desired status from the *previous* reconcile.

### Motivation

A common e2e flake: after `patch_custom_resource(...)`, tests do `wait_on_condition(ref, ACK.ResourceSynced, "True")` and immediately assert on status. But ACK re-affirms `ACK.ResourceSynced=True` after every successful reconcile, so right after a patch the condition can still read `True` from the reconcile that ran *before* the patch. The wait becomes a no-op and the test reads stale status.

Generation-based checks don't help: ACK records no observed generation (the core `Condition` type has no `observedGeneration`), and `metadata.generation` is bumped by the API server the instant the spec is patched, before the controller observes it.

What does work: ACK's `SetSynced` rewrites `lastTransitionTime` to "now" on **every** reconcile (not only on status changes). So capturing that timestamp before the patch and waiting for a strictly newer one reliably detects that the controller reconciled the change.

### Changes

- `resource.wait_on_condition_after(...)` — a **new** function (not a change to `wait_on_condition`, so existing callers are unaffected) that treats a condition as met only once its `lastTransitionTime` is strictly newer than a supplied `last_transition_after`. When `last_transition_after` is `None` it behaves like `wait_on_condition`.
- `resource.parse_condition_last_transition_time(condition)` — parses a condition's `lastTransitionTime` into a timezone-aware `datetime`.
- `condition.get_synced_last_transition_time(ref)` — captures the `ACK.ResourceSynced` condition's `lastTransitionTime` for the pattern above.

### Usage

```python
synced_at = condition.get_synced_last_transition_time(ref)
k8s.patch_custom_resource(ref, updates)
assert k8s.wait_on_condition_after(
    ref, condition.CONDITION_TYPE_RESOURCE_SYNCED, "True",
    last_transition_after=synced_at,
)
```

Capture the timestamp **before** the patch: it gives a stable baseline that the patch's reconcile is guaranteed to exceed. (Capturing after the patch races the controller and can skip the very reconcile that processed the change.)

### Related

First consumer: aws-controllers-k8s/firehose-controller#24, which replaces fixed `time.sleep(...)` waits in the delivery stream e2e tests with this pattern.

### Testing

`py_compile` passes on both modified modules. Existing `wait_on_condition` behavior is unchanged (new function is additive).

### DIFF
--- a/src/acktest/k8s/condition.py
+++ b/src/acktest/k8s/condition.py
@@ -15,6 +15,9 @@
 
 import pytest
 
+from datetime import datetime
+from typing import Optional
+
 from . import resource
 
 CONDITION_TYPE_ADOPTED = "ACK.Adopted"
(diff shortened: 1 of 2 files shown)


sapphirew (approved):
/lgtm

PR #1014, opened 2026-07-16
fix: refresh assumed-role creds before helm test phase

michaelhtm:
Description of changes:
The code-generator core-validator runs unit, e2e and helm phases back-to-back
in one Prow pod. wrapper.sh mints a single 1-hour assume-role credential at job
start, so the helm phase -- starting after a long e2e phase -- inherits an
expired credential and fails with ExpiredToken. (Service repos don't hit this:
they run each phase as a separate pod with its own fresh credential.)

Preserve the Pod Identity reference in wrapper.sh and re-mint the credential
from it at the start of the helm phase. Pod Identity is reachable only from the
pod and does not expire, so the refresh always starts from a valid source.


sapphirew:
Would this change fix the same issue? https://github.com/aws-controllers-k8s/test-infra/commit/de3d4d970a1f6ae0b8ddf6c695e59b90446435a3. Can you bump the acktest version for the controller you are working on?

sapphirew (approved):
/lgtm

PR #1017, opened 2026-07-16
fix: keep ${PROW_IMAGES_REPO_URI} variable in generated prow manifests

naclonts:
The `build-prow-images` bot resolves `image_repo` to a hardcoded ECR URI before generating `jobs.yaml`/`job-config-job.yaml`, so every `make prow-gen` run produces ~1000 lines of churn flipping between the variable and the literal URI.

Split the config inputs: build/push keeps the resolved config (real ECR URI, needed for buildah push), while generation uses a new `--images-generate-config-path` pointing at the raw config so manifests retain `${PROW_IMAGES_REPO_URI}` (matching `make prow-gen`). Mirrors the existing plugins path. The flag defaults to `--images-config-path` when unset, so behavior is unchanged if not passed.


knottnt (approved):
/lgtm

PR #1031, opened 2026-07-30
 Add a dedicated EKS build cluster for Prow pre-submit jobs

knottnt:
Issue #, if available:

Description of changes:
Provisions a dedicated EKS build cluster to run Prow pre-submit jobs in isolation from the Prow control plane, and wires up the control-plane machinery to use it.

This PR provisions and stages everything but leaves it dormant — the control plane is untouched and no jobs are rerouted. It ships in a state where the build cluster comes up, Flux stages its connection details and RBAC, but the four control-plane components render byte-identical to today. Two small follow-ups flip it live (see Rollout below).

What's included

  1. Provision the build cluster via ACK — a dedicated EKS Auto Mode cluster (own VPC on a non-overlapping 10.1.0.0/16, subnets, NAT, IAM roles) created entirely through ACK controllers running on the control-plane cluster — no
  Terraform. A single Flux Kustomization reconciles the whole tree; ordering is handled by ACK reference resolution. Includes the cluster's EKS access entries and pod identity associations. The capability role is widened (EC2 VPC
  permissions + ${STACK_NAME}-*cluster EKS scope) so ACK can create it.
  2. Wire Flux to reach the build cluster — the control-plane kustomize-controller reaches the build cluster's API via workload identity (a provider: aws kubeconfig ConfigMap backed by a Pod Identity holding eks:DescribeCluster) and
  remote-applies the in-cluster objects ACK can't create there: the test-pods namespace, the pre-submit service account, and the prow-build-cluster RBAC. A connection Job reads the cluster's endpoint/CA from its ACK Cluster CR status into a ConfigMap; prow-charts depends on that Job (wait: true) so the components' kubeconfig isn't rendered until the cluster is ACTIVE.
  3. Wire the four control-plane components — mount a build-cluster kubeconfig into prow-controller-manager, sinker, crier, and deck so they load the build cluster as an additional Prow build cluster. The build context authenticates
  with an aws-iam-authenticator exec credential (already in the upstream base images) as the components' own Pod Identity, which the cluster's access entry authorizes. Gated on buildCluster.enabled, which ships false — the Deployments render unchanged while disabled.
  4. Add dormant pre-submit routing — a presubmit_cluster option in the jobs generator that, when set, emits a cluster: field on every pre-submit. Left empty, so the regenerated jobs.yaml is behaviorally unchanged (pre-submits still run
  on the in-cluster default). Post-submits and periodics always stay on default.

  Safety / blast radius

  - buildCluster.enabled: false + presubmit_cluster: "" → this PR only adds objects; the running control plane is not modified and no jobs move.
  - Flux ordering (wait: true + dependsOn) means the components never receive a build-cluster kubeconfig until the cluster is ACTIVE and its endpoint/CA exist.

  Rollout (follow-ups after this merges & is verified)

  1. Flip buildCluster.enabled → true in flux/prow/charts/prow-config.yaml — connects the four components to the build cluster (rolling restart of those four; no job impact). Verify they pick up the build context.
  2. Set presubmit_cluster: "build" in prow/jobs/jobs_config.yaml + make prow-gen — routes pre-submits to the build cluster.

  Rollback at any stage is a one-line revert of the corresponding flag.


flux/ack/build-cluster/infra/cluster.yaml
```diff
@@ -0,0 +1,55 @@
...
+        name: build-subnet-private-b
+    endpointPublicAccess: true
+    endpointPrivateAccess: true
+  accessConfig:
+    authenticationMode: API
+    bootstrapClusterCreatorAdminPermissions: true
+  bootstrapSelfManagedAddons: false
+  computeConfig:
+    enabled: true
+    nodePools:
```
gustavodiaz7722: Q: Is general purpose sufficient for our use case?

flux/ack/build-cluster/infra/cluster.yaml
```diff
@@ -0,0 +1,55 @@
...
+    subnetRefs:
+    - from:
+        name: build-subnet-private-a
+    - from:
+        name: build-subnet-private-b
+    endpointPublicAccess: true
+    endpointPrivateAccess: true
+  accessConfig:
+    authenticationMode: API
+    bootstrapClusterCreatorAdminPermissions: true
```
gustavodiaz7722: Q: is this needed? Can we set CreatorAdminPermissions to false instead?

gustavodiaz7722:
nit: update flux/README.md, flux/ack/README.md to reflect change in prow infra.

flux/prow/build-cluster-connection/job.yaml
```diff
@@ -0,0 +1,60 @@
...
+        command:
+        - /bin/bash
+        - -c
+        - |
+          set -euo pipefail
+
+          CR_NS="ack-system"
+          CR="build-cluster"
+
+          echo "Reading connection details from Cluster CR ${CR_NS}/${CR}..."
```
gustavodiaz7722: Q: does this log line print correctly? I think I saw an issue before where we needed double escape, similar to prow/jobs/job-config-job.yaml escape as $${VAR}.

flux/ack/capability/role/ack-capability-role.yaml
```diff
@@ -35,15 +35,63 @@ spec:
...
+              "ec2:DisassociateRouteTable",
+              "ec2:DescribeManagedPrefixLists",
+              "ec2:DescribeSecurityGroups",
+              "ec2:DescribeSecurityGroupRules",
+              "ec2:RevokeSecurityGroupEgress",
+              "ec2:RevokeSecurityGroupIngress",
+              "ec2:CreateTags",
+              "ec2:DeleteTags"
+            ],
+            "Resource": "*"
```
gustavodiaz7722, line 91: Would it be possible to scope this to specific resources?

flux/ack/build-cluster/infra/cluster.yaml
knottnt: I think this can be set to false.

flux/ack/build-cluster/infra/cluster.yaml
knottnt: Hmm, thinking about it would make sense to apply a similar NodePool to the one currently be used for prow-compute. Will update.

flux/prow/build-cluster-connection/job.yaml
knottnt: Good catch! Will fix.

PR #1037, opened 2026-08-06
Patch OS CVEs in Prow and job images to clear container findings

gustavodiaz7722:
## Summary

Image scanning flags the Prow control plane, our own CI job images, and the secrets-store CSI driver for outdated OS packages — `expat`, `openssl` and `curl`. None of these are application defects, and none are fixed by moving to a newer upstream version, because the outdated packages are inherited from pinned base images.

**Why a version bump doesn't fix this.** `kubernetes-sigs/prow` pins its ko base images in [`.ko.yaml`](https://github.com/kubernetes-sigs/prow/blob/main/.ko.yaml) and refreshes that pin every 6–12 months. It currently points at `gcr.io/k8s-staging-test-infra/git-custom-k8s-auth:v20260512-1f34ef0df3`, last bumped 2026-06-09, while `v20260729-*` bases already exist upstream. The tag we run, `v20260702-fff8399c7`, was built *after* that bump, so it already carries the newest base upstream offers. The `eks-distro-build-tooling` Prow images don't help either — [their build](https://github.com/aws/eks-distro-build-tooling/blob/main/projects/kubernetes/test-infra/Makefile) sets `BASE_IMAGE` to the same upstream image and adds no OS updates.

So the fix has to apply OS updates ourselves.

### 1. Bump the secrets-store CSI driver addon

`aws-secrets-store-csi-driver-provider` `v3.1.1-eksbuild.1` → `v3.1.2-eksbuild.1`. The addon bundles the `secrets-store-csi-driver` image that the DaemonSet runs, so the bundle version is what governs how current those packages are.

### 2. Apply OS updates in ACK-owned images

The final stage of all 18 Dockerfiles now runs `apt-get upgrade` / `apk -U upgrade` / `dnf upgrade` as appropriate, and the tags in the three `images_config.yaml` files are bumped so `build-prow-images` rebuilds and republishes them.

Per the repo's two-phase flow this commit bumps **only** the image config. The regenerated `jobs.yaml` / `job-config-job.yaml` / `agent-workflows.yaml` / `prow/plugins/deployments/` land in the follow-up bot PR, once the images are actually in ECR. The one exception kept here is the `${PROW_IMAGES_REPO_URI}` un-hardcoding in `job-config-job.yaml`, which reverses a literal registry baked in by #1008; its tag stays at the image that exists today.

The invariant: **every image tag referenced by a file this PR applies directly is already in ECR.**

### 3. Rebuild instead of straight-copying mirrored Prow images

`prow-mirror` previously ran `crane copy`. It now uses buildah to pull each image, run `apk upgrade`, and publish it under two tags:

| Tag | Role |
|---|---|
| `<upstream-tag>-ack.<PROW_PATCH_REVISION>` | Authoritative. What `prow-config` pins, so a revision bump is what drives the rollout. |
| `<upstream-tag>` | Compatibility alias, same patched content. A reference that hasn't moved to the suffixed tag degrades to *patched but unpinned* rather than *silently unpatched*. |

The Kustomization gains `wait: true` + `timeout: 30m` so `prow-charts`, which depends on it, cannot roll out the suffixed tags before the Job has produced them — `dependsOn` alone gates on apply, not completion.

Two safety properties:
- **Idempotent.** An init container skips any image whose patched tag is already in ECR, so the hourly reconcile is a fast no-op once everything is built.
- **Fails loudly.** If a base image has no `apk` (e.g. upstream moves a component to distroless) the job refuses to publish rather than silently shipping it unpatched.

`PROW_PATCH_REVISION` decouples our patch cadence from upstream's base-image cadence. It's a **monotonic counter, never reset**: `PROW_VERSION` and `TOOLS_VERSION` move independently but share the one suffix, so reusing a revision would point the untouched tag at an image that already exists in ECR — the mirror would skip the rebuild and roll that component back to its least-patched build. `upgrade-prow.sh --bump-patch` rebuilds the pinned version against newly published OS updates.

## Testing

Local only; nothing applied to a cluster.

- `kustomize build` succeeds for all 7 touched kustomizations.
- Both embedded mirror-job scripts extracted from the rendered Job, put through Flux's substitution rules, and syntax-checked (`sh -n`, `bash -n`). `bash -n` on `upgrade-prow.sh`.
- Mirror job logic exercised against stubbed `aws`/`buildah`:
  - cold cache → all 15 images queued with correct source registries and tags;
  - warm cache → 0 queued, exits 0 without building;
  - transient push failure → retried and succeeded;
  - base without `apk` → refused, **no push attempted**, no pointless retries, exits 1;
  - dual-tag → two pushes and one `tag` alias per image, suffixed and bare.
- `helm template prow/config` with the real `HelmRelease` values renders all 9 component and 4 utility images on the `-ack.1` tag; 0 left on an unpinned bare tag.
- Generator output confirmed byte-identical to `main` for all five generated files; `job-config-job.yaml` differs by the single registry line.
- Enumerated every `prow-*` tag referenced in directly-applied files — all are pre-existing tags, none of the bumped ones.
- `make prow-gen` re-run to confirm the *sources* generate cleanly against #1031's build-cluster changes; output intentionally not committed.
- No IAM changes needed: `${STACK_NAME}-prow-mirror-role` already holds `ecr:*` on `repository/prow/*`.

## Reviewer notes

- **The mirror Job runs privileged.** buildah needs it to unpack and re-commit layers. This matches the existing `build-prow-images` postsubmit, which already runs `privileged: true` in `test-pods`. Worth a conscious ack.
- **Addon bundle contents unverified.** The EKS addon image registry doesn't grant `ecr:DescribeImages` to arbitrary accounts, so I couldn't confirm which `secrets-store-csi-driver` tag `v3.1.2-eksbuild.1` ships. Worth confirming with a scan after rollout.
- **`Dockerfile.docs` is on `alpine:3.17.2`, which is EOL**, so `apk upgrade` is a no-op there for most advisories. Flagged in a comment, left for a follow-up that revalidates the docs build on a newer Python.
- **Two images are unpinned** (`alpine:latest` in `scan-controllers-cve` and `upgrade-go-version`). Left as-is to keep this diff scoped.

## Follow-ups

- `flux/prow/build-cluster-connection/job.yaml` pins `prow-kubectl-0.0.2` and is not in the generator's output set, so it needs a manual bump once `0.0.3` is in ECR. Documented in `bootstrap/README.md` under "Exceptions the generator does not cover".
- The Flux controller images come through the ECR pull-through cache, which is passthrough-only and can't carry a patch layer. Needs either a Flux upgrade (currently 2.18.4) or dropping the pull-through cache in favour of mirror-and-patch like Prow.
- A PR to `kubernetes-sigs/prow` bumping `.ko.yaml` to the `v20260729-*` bases would reduce recurrence for everyone.

### DIFF
--- a/bootstrap/README.md
+++ b/bootstrap/README.md
@@ -141,8 +141,19 @@ git push
 
 ## Upgrading Prow
 
-Prow images are mirrored from upstream into a private ECR registry. The version
-is controlled by a single ConfigMap (`flux/prow/version/prow-version-configmap.yaml`).
+Prow images are mirrored from upstream into a private ECR registry **and rebuilt
+with current OS packages** on the way through. Upstream Prow pins its ko base
+images in `.ko.yaml` and only refreshes that pin every few months, so even the
+newest Prow release ships stale `expat`/`openssl`/`curl`. The `prow-mirror` Job
+runs `apk upgrade` over each image and publishes it as
+`<upstream-tag>-ack.<PROW_PATCH_REVISION>`, plus the bare `<upstream-tag>` as a
+compatibility alias carrying the same patched content. The suffixed tag is what
+`prow-config` pins, so bumping the revision is what triggers a rollout; the bare
+alias means a reference that has not moved to the suffixed tag still gets patched
+packages rather than silently staying vulnerable.
+
+Both the version and the patch revision live in a single ConfigMap
+(`flux/prow/version/prow-version-configmap.yaml`).
 
 ```bash
 # Auto-detect latest tags for both Prow core and tools, update CRD
@@ -154,16 +165,74 @@ is controlled by a single ConfigMap (`flux/prow/version/prow-version-configmap.y
 # Preview changes without modifying files
 ./scripts/upgrade-prow.sh --dry-run
 
+# Re-patch the CURRENT Prow version against newly published OS security
+# updates, without moving to a new upstream release. Increments
+# PROW_PATCH_REVISION, which is what makes the mirror Job rebuild.
+./scripts/upgrade-prow.sh --bump-patch
+
 # Commit and push
 git add flux/prow/ prow/config/
 git commit -m "chore(prow): upgrade to <tag>"
 git push
-# Flux reconciles: mirror job copies new images to ECR, then Prow redeploys
+# Flux reconciles: mirror job rebuilds patched images into ECR, then Prow redeploys
 ```
+
+`PROW_PATCH_REVISION` is a monotonic counter and is never reset. `PROW_VERSION`
+and `TOOLS_VERSION` move independently but share the one suffix, so resetting it
+on a version change would repoint the untouched tag at a revision that already
+exists in ECR — the mirror would skip the rebuild and that component would
+silently roll back to its least-patched build.
 
 Image sources:
 - Prow core: `us-docker.pkg.dev/k8s-infra-prow/images` (13 images)
 - Tools (`label_sync`, `commenter`): `gcr.io/k8s-staging-test-infra`
+
+The mirror Job refuses to publish an image whose base has no `apk`, rather than
+silently shipping it unpatched. If upstream moves a component to a distroless
+base, that component will fail loudly and needs a different patch strategy.
+
+### Reacting to an image vulnerability report
+
+**Mirrored Prow components** (`.../prow/<component>`) are almost always outdated
+OS packages in the upstream base image, not a Prow defect. Bumping the Prow
+version alone usually does **not** fix them. Run
+`./scripts/upgrade-prow.sh --bump-patch`, push, then force-reconcile
+`prow-mirror`.
+
+**Our own job images** (`public.ecr.aws/<alias>/...-prow-images:*`) apply OS
+updates at build time, so they only need a tag bump to be rebuilt. Note the
+two-phase flow — do not shortcut it:
+
+1. In your PR, bump the tag in `prow/jobs/images_config.yaml` (or the
+   `prow/plugins/` / `prow/agent-workflows/` equivalent) and **nothing else**.
+   `compareImageVersions` only builds a tag strictly greater than what is in
+   ECR, so an unchanged tag is never rebuilt.
+2. On merge, the `build-prow-images` postsubmit builds and pushes the new tags.
+3. A bot PR then commits the regenerated `jobs.yaml` / `job-config-job.yaml` /
+   `agent-workflows.yaml` / `prow/plugins/deployments/`.
+
+Do not run `make prow-gen` and commit its output in step 1. `job-config-job.yaml`
+is applied directly by the `prow-jobs` Kustomization (`interval: 5m`,
+`force: true`), so committing a tag before the image exists makes Flux recreate
+`job-config-substitutor` on a missing image within minutes of merge, halting
+`job-config` refreshes until the postsubmit lands.
+
+#### Exceptions the generator does not cover
+
+- **`build-prow-images` is safe to bump on its own.** It is the image the
+  postsubmit itself runs as, but the running pod resolves it from the
+  already-applied `job-config` ConfigMap, not from `images_config.yaml`. So the
+  job builds its own successor. This only holds while step 1 above is respected;
+  hand-committing a regenerated `jobs.yaml` would point the job at the image it
+  has not built yet.
+- **`flux/prow/build-cluster-connection/job.yaml`** pins a `prow-kubectl` tag
+  but is not in the generator's output set, so `make prow-gen` will never update
+  it. Bump it by hand once the new tag is in ECR, or its OS packages silently
+  stay behind.
+
+Keep notes like these here rather than as comments in `images_config.yaml`: the
+`upgrade-go-version` periodic rewrites that file through a typed struct and a
+YAML encoder, which strips comments.
 
 ## Re-running Terraform
 
@@ -234,6 +303,8 @@ Common kustomization names:
 | `ack-prow` | Prow AWS resources (S3, Route53) |
 | `ack-flux` | ECR pull-through cache |
 | `prow-crds` | Prow CRDs |
+| `prow-version` | `prow-version` ConfigMap (Prow/tools versions + patch revision) |
+| `prow-mirror` | Job that rebuilds upstream Prow images with current OS packages |
 | `prow-charts` | Prow Helm releases |
 
 If a kustomization shows `dependency '<name>' is not ready`, trigger the
--- a/flux/ack/cluster/addons/addons.yaml
+++ b/flux/ack/cluster/addons/addons.yaml
@@ -8,7 +8,13 @@ metadata:
 spec:
   clusterName: ${STACK_NAME}-cluster
   name: aws-secrets-store-csi-driver-provider
-  addonVersion: v3.1.1-eksbuild.1
+  # Keep this at the newest published version. The addon bundles the upstream
+  # secrets-store-csi-driver image, so an old bundle leaves the driver DaemonSet
+  # running outdated OS packages. Check for newer versions with:
+  #   aws eks describe-addon-versions \
+  #     --addon-name aws-secrets-store-csi-driver-provider \
+  #     --kubernetes-version 1.35 --region us-west-2
+  addonVersion: v3.1.2-eksbuild.1
   configurationValues: |-
     {
       "secrets-store-csi-driver": {
--- a/flux/prow.yaml
+++ b/flux/prow.yaml
@@ -57,6 +57,14 @@ spec:
   path: ./flux/prow/mirror
   prune: false
   force: true
+  # Wait for the Job to COMPLETE, not just apply. prow-charts depends on this
+  # Kustomization and pins the -ack.<revision> tags this Job produces, so
+  # without wait the components roll out before the tags exist and land in
+  # ImagePullBackOff. A serial buildah pull/upgrade/commit/push across 15 images
+  # on vfs storage takes minutes rather than the seconds `crane copy` took, so
+  # the timeout is deliberately generous.
+  wait: true
+  timeout: 30m
   postBuild:
     substituteFrom:
     - kind: ConfigMap
--- a/flux/prow/charts/prow-config.yaml
+++ b/flux/prow/charts/prow-config.yaml
@@ -52,47 +52,47 @@ spec:
       server: "${BUILD_CLUSTER_ENDPOINT}"
       certificateAuthorityData: "${BUILD_CLUSTER_CA}"
     crier:
-      image: "${ACCOUNT_ID}.dkr.ecr.${REGION}.amazonaws.com/prow/crier:${PROW_VERSION}"
+      image: "${ACCOUNT_ID}.dkr.ecr.${REGION}.amazonaws.com/prow/crier:${PROW_VERSION}-ack.${PROW_PATCH_REVISION}"
       serviceAccount:
         name: "prow-deployment-service-account"
     deck:
-      image: "${ACCOUNT_ID}.dkr.ecr.${REGION}.amazonaws.com/prow/deck:${PROW_VERSION}"
+      image: "${ACCOUNT_ID}.dkr.ecr.${REGION}.amazonaws.com/prow/deck:${PROW_VERSION}-ack.${PROW_PATCH_REVISION}"
       serviceAccount:
         name: "prow-deployment-service-account"
     ghproxy:
-      image: "${ACCOUNT_ID}.dkr.ecr.${REGION}.amazonaws.com/prow/ghproxy:${PROW_VERSION}"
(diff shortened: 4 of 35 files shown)


sapphirew:
### Code review

Found 6 issues:

1. `Dockerfile.build-prow-images` gets `dnf -y upgrade --refresh` but its tag stays at `0.0.71`, so the patch is never built. [`compareImageVersions`](https://github.com/aws-controllers-k8s/test-infra/blob/247826fa462a706b7e3d8bc6b2331ff1e068474b/prow/jobs/tools/cmd/command/patch_prowjobs_helper.go#L118-L146) only builds a tag strictly greater than what is in ECR, so the one image that runs `privileged: true` is the only one in the fleet that keeps shipping unpatched OS packages. Note the NOTE's premise looks inverted: [#948](https://github.com/aws-controllers-k8s/test-infra/pull/948) and [#991](https://github.com/aws-controllers-k8s/test-infra/pull/991) each bumped this tag alone, because the running job's pod image comes from the already-applied `job-config` (`jobs.yaml:40021` pins `0.0.71`, which exists) — not from `images_config.yaml`. The bump appears safe as long as `jobs.yaml` is left to the follow-up bot PR (see issue 3).

https://github.com/aws-controllers-k8s/test-infra/blob/247826fa462a706b7e3d8bc6b2331ff1e068474b/prow/jobs/images_config.yaml#L4-L10

2. A single `PROW_PATCH_REVISION` is shared by two independently-moving tags, so a `TOOLS_VERSION`-only bump rolls Prow core back to its least-patched build. `VERSION_CHANGED=true` is set when *either* `PROW_VERSION` or `TOOLS_VERSION` moves, then the revision resets to `1` unconditionally. From steady state `PROW_VERSION=X, PROW_PATCH_REVISION=3`, a run where only the tools tag moved repoints all 13 components at `X-ack.1` — a tag that already exists in ECR, so the mirror's plan container skips the rebuild and the control plane silently reverts, undoing exactly the remediation this PR delivers. The symmetric case strands `label_sync`/`commenter`. Needs per-tag revisions or a monotonic counter.

https://github.com/aws-controllers-k8s/test-infra/blob/247826fa462a706b7e3d8bc6b2331ff1e068474b/scripts/upgrade-prow.sh#L196-L201

3. Committing the regenerated `jobs.yaml`/`job-config-job.yaml` alongside the `images_config.yaml` bumps inverts the repo's two-phase flow and breaks the substitutor Job on merge. Historically human PRs touch only `images_config.yaml` ([#1015](https://github.com/aws-controllers-k8s/test-infra/pull/1015), [#951](https://github.com/aws-controllers-k8s/test-infra/pull/951), [#1009](https://github.com/aws-controllers-k8s/test-infra/pull/1009)) and a bot PR commits the regenerated output *after* the images are pushed ([#1008](https://github.com/aws-controllers-k8s/test-infra/pull/1008), [#1016](https://github.com/aws-controllers-k8s/test-infra/pull/1016)). `job-config-job.yaml` is applied directly by the `prow-jobs` Kustomization (`interval: 5m`, `force: true`), so within ~5 min of merge Flux recreates `job-config-substitutor` on a `prow-kubectl-0.0.3` that does not exist yet, halting `job-config` refreshes until the postsubmit lands. Keep the `${PROW_IMAGES_REPO_URI}` fix here (it un-hardcodes the registry baked in by #1008) and drop the premature tag.

https://github.com/aws-controllers-k8s/test-infra/blob/247826fa462a706b7e3d8bc6b2331ff1e068474b/prow/jobs/job-config-job.yaml#L139-L141

4. `flux/prow/build-cluster-connection/job.yaml` is stranded on the unpatched `prow-kubectl-0.0.2`. `Dockerfile.kubectl` was patched and the tag moved to `0.0.3` everywhere the generator writes, but this file is not in the generator's output set, so `make prow-gen` will never reach it and the CVE fix silently misses this Job.

https://github.com/aws-controllers-k8s/test-infra/blob/247826fa462a706b7e3d8bc6b2331ff1e068474b/flux/prow/build-cluster-connection/job.yaml#L23-L25

5. `prow-mirror` has no `wait: true`, so `prow-charts` rolls out `-ack.N` tags before they exist. `dependsOn` gates on *apply*, not completion. With `crane copy` that window was seconds; a serial buildah pull/upgrade/commit/push over 15 images on `vfs` makes it many minutes, and since `PROW_PATCH_REVISION: "1"` is brand new no `-ack.1` tag exists anywhere yet — so the first reconcile after merge reliably puts deck/hook/tide/crier into `ImagePullBackOff`. The sibling `prow-build-cluster-connection` sets `wait: true` + `timeout: 5m` for precisely this reason.

https://github.com/aws-controllers-k8s/test-infra/blob/247826fa462a706b7e3d8bc6b2331ff1e068474b/flux/prow.yaml#L52-L60

6. The new NOTE in `images_config.yaml` will be silently deleted by the `upgrade-go-version` periodic. [`patchImageConfigVersionFile`](https://github.com/aws-controllers-k8s/test-infra/blob/247826fa462a706b7e3d8bc6b2331ff1e068474b/prow/jobs/tools/cmd/command/helpers.go#L71-L80) unmarshals into a typed struct with a plain `map[string]string` and re-encodes with `yaml.NewEncoder`, which drops comments — so the only record of the builder constraint disappears on the next Go bump. Relatedly, `bootstrap/README.md:197-201` gives the general "bump the tag, run `make prow-gen`, let the postsubmit rebuild" recipe without the builder exception, which is the one image where running `prow-gen` is what causes the problem. Both belong in the README.

https://github.com/aws-controllers-k8s/test-infra/blob/247826fa462a706b7e3d8bc6b2331ff1e068474b/prow/jobs/images_config.yaml#L4-L10

Verified clean: `$$`-escaping throughout the new mirror shell (the [#1031](https://github.com/aws-controllers-k8s/test-infra/pull/1031) double-escape trap); `apk`/`apt`/`dnf` upgrades all land in stages that reach the final image (no discarded builder stages across all 21 Dockerfiles); the 13 suffixed refs in `prow-config.yaml` are set-identical to what the mirror builds; the "fails loudly on unpatchable base" rc=2 path; and generated output matches templates tag-for-tag.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

gustavodiaz7722:
Thanks — all six were real. Fixed in `0cbb3f6`. Verified each claim against the code before changing anything; every one held up.

### 1. Builder tag never rebuilt

Agreed, and my `NOTE` had the premise backwards. Confirmed `compareImageVersions` gates on `isGreaterVersion(latestTag, ecrTag)`, strictly greater, so `0.0.71` would never have been rebuilt — the one `privileged: true` image in the fleet would have kept shipping unpatched packages. And confirmed the running pod resolves its image from the applied `job-config` (`jobs.yaml:40021` pins `0.0.71`, which exists), not from `images_config.yaml`, so the job does build its own successor.

Bumped to `prow-build-prow-images-0.0.72`. Also re-checked that the `regex()` in `bootstrap/images.tf` and the `grep | awk` in `bootstrap-images.sh` still resolve the tag after removing the comment.

### 2. Shared `PROW_PATCH_REVISION`

Fixed with a monotonic counter that is never reset. I walked your failure case: from `PROW_VERSION=X, PROW_PATCH_REVISION=3`, a tools-only bump set the revision back to `1`, repointing all 13 components at `X-ack.1` — which already exists, so the plan container skips the rebuild and the control plane silently reverts. Exactly the remediation this PR is supposed to deliver, undone.

It now increments whenever either tag moves or `--bump-patch` is passed. I chose the monotonic counter over per-tag revisions because it always yields a tag that does not exist yet, so there is no path where a stale image is silently reused. The cost is rebuilding images whose upstream tag did not move, which is a fresh patch rather than a regression. Documented in the ConfigMap and the README.

### 3. Two-phase flow

Reverted all generator output to `upstream/main`: `jobs.yaml`, `labels.yaml`, `agent-workflows.yaml`, `prow/plugins/deployments/*`. Kept the `${PROW_IMAGES_REPO_URI}` un-hardcoding in `job-config-job.yaml` and reverted the tag to `0.0.2` as you suggested — that one line is now the only delta in that file. Also confirmed nothing in CI enforces `prow-gen` drift, so the lag is safe.

The invariant now is: **every image tag referenced by a file this PR applies directly is a tag already in ECR.** There is a check for it in the testing notes below.

One knock-on worth flagging: reverting `jobs.yaml` leaves the periodics on the bare tools tag while the mirror publishes only `-ack.N`, which would have broken `label_sync`/`commenter` until the bot PR. Rather than special-case it, the mirror now publishes each patched image under **both** the suffixed tag and the bare upstream tag. The suffixed tag stays authoritative and is what `prow-config` pins, so a revision bump is still what drives the rollout; the bare tag becomes a compatibility alias carrying the same patched content. That also means a reference that has not moved to the suffixed tag degrades to "patched but unpinned" instead of "silently unpatched" — which is the failure mode in your issue 4.

### 4. `build-cluster-connection` stranded on `prow-kubectl-0.0.2`

Confirmed it is not in the generator's output set, so `make prow-gen` will never reach it. I have deliberately **left it at `0.0.2`** to preserve the invariant in 3 — bumping it now would reference an image that does not exist until the postsubmit runs, and this is a `prow-kubectl` job image, so the mirror's dual-tag does not cover it.

It needs a manual bump once `0.0.3` is in ECR. Rather than leave that to memory I added it to `bootstrap/README.md` under "Exceptions the generator does not cover". Happy to own the follow-up bump alongside the bot PR, or to fold it in here if you would rather accept the short window — the build cluster is still gated off (`buildCluster.enabled: false`), so the impact of a transient pull failure there is low. Your call.

### 5. `prow-mirror` missing `wait: true`

Added `wait: true` and `timeout: 30m`, with a comment pointing at the same reasoning as the sibling `prow-build-cluster-connection`. Used 30m rather than 5m because 15 serial buildah pull/upgrade/commit/push cycles on `vfs` is minutes, not the seconds `crane copy` took.

### 6. Comment stripped by the `upgrade-go-version` periodic

Confirmed — `patchImageConfigVersionFile` does `os.Create` then `yaml.NewEncoder().Encode()` over a typed struct with a plain `map[string]string`, so comments do not survive. Removed the comment from `images_config.yaml` and moved everything into `bootstrap/README.md`:

- the two-phase flow, with an explicit "do not run `make prow-gen` and commit its output" warning and the reason (`prow-jobs` at `interval: 5m`, `force: true`)
- the `build-prow-images` exception and why it is safe to bump alone
- the manifests the generator does not cover
- a note to keep such documentation out of `images_config.yaml` for exactly this reason

You were also right that the README recipe gave the general "bump, `prow-gen`, let the postsubmit rebuild" flow without the builder exception. That section is rewritten.

### Testing

- Diff shrank from 40 files / +1477 −1180 to 35 / +485 −117, entirely from dropping the generated churn.
- Generator output confirmed byte-identical to `upstream/main` for all five files; `job-config-job.yaml` differs by the single registry line.
- Enumerated every `prow-*` tag referenced in the files this PR applies directly — all 15 are the pre-existing tags, none of the bumped ones.
- Dual-tag exercised against a stubbed `buildah`: two pushes and one `tag` alias per image, suffixed and bare, correct registries.
- Re-ran the earlier matrix after the changes: cold cache queues 15 with correct source registries, warm cache queues 0 and exits 0, transient push failure retries then succeeds, and a base without `apk` refuses with zero pushes and no pointless retries.
- `kustomize build` on all 7 touched kustomizations; both embedded mirror scripts pass `sh -n` / `bash -n` after Flux substitution; `bash -n` on `upgrade-prow.sh`; all changed YAML parses.

Your "verified clean" list still holds — the `$$`-escaping, the upgrade landing in stages that reach the final image, the suffixed refs in `prow-config.yaml` being set-identical to what the mirror builds, and the rc=2 refuse path are all unchanged by this round.

sapphirew:
Non-blocking follow-up on the tag bumps: the `kubectl` bump orphans one consumer.

[`flux/prow/build-cluster-connection/job.yaml:24`](https://github.com/aws-controllers-k8s/test-infra/blob/0cbb3f626bc60ff9b342968cec53087678c5e7ee/flux/prow/build-cluster-connection/job.yaml#L23-L25) hardcodes `${PROW_IMAGES_REPO_URI}:prow-kubectl-0.0.2`, while [`prow/jobs/images_config.yaml`](https://github.com/aws-controllers-k8s/test-infra/blob/0cbb3f626bc60ff9b342968cec53087678c5e7ee/prow/jobs/images_config.yaml#L10-L12) now declares `prow-kubectl-0.0.3`. Both said `0.0.2` before this PR, so the bump is what splits them.

That file isn't in the generator's output set, so `make prow-gen` won't rewrite it and the "Patch Prow Image Versions" bot won't either — it needs a manual edit whenever the kubectl tag moves. Latent rather than urgent: it keeps working as long as `0.0.2` stays in ECR.

(`prow/jobs/job-config-job.yaml:137` is also still on `0.0.2`, but that one is correct — the bot regenerates it, which is the two-phase flow this PR restores.)

PR #1059, opened 2026-08-20
Disable Flux prune ahead of the Argo CD migration (stage 1/8)

gustavodiaz7722:
**Stage 1 of 8** in the Flux → Argo CD migration. See `docs/argocd-migration-runbook.md`.

This stage only makes objects harder to delete. It changes no behaviour: Flux keeps
reconciling exactly as before, and every path keeps rendering the same content.

## What this does

- `prune: false` on all 19 Kustomizations that had it enabled
- `kustomize.toolkit.fluxcd.io/prune: disabled` and `helm.sh/resource-policy: keep` on the
  objects those paths own, so deletion is structurally prevented rather than only configured
  away
- `enableNetworkAddressUsageMetrics: false` on the build VPC, unrelated but trivial and already
  in this range

## Why it comes first

Every later stage suspends or deletes something, and **a Kustomization with prune enabled
garbage-collects its inventory when it is deleted.** In staging `prow-charts` held the
`prow-config` HelmRelease, whose Helm release owns Prow's nine Deployments — so pruning it would
have taken the control plane down. Disabling prune everywhere first is what makes the whole
migration reversible up to the cutover.

`prune: false` is also why the migration leaves debris that has to be swept by hand at each
stage. That is a deliberate trade: reversibility over tidiness.

## Two things a reviewer should know

**`prometheus` and `prometheus-dashboards` deliberately keep `prune: true`.** They are never
converted to charts — a later stage deletes them outright — so there is no window in which prune
could delete a stale object, and prune reclaiming their inventory on deletion is the wanted
outcome rather than the hazard.

**The fourth commit is not a cherry-pick.** `prow-agent-workflows`, `prow-crds`, `prow-plugins`
and `secrets` had prune disabled in staging by commits belonging to later stages, so delivering
this stage on its own would have left those four at `prune: true` — precisely the condition this
stage exists to prevent. It was written for this PR.

## Testing

- Every changed YAML file parses (35 files)
- Diff contains nothing but the prune flips, the two annotations, and the one metrics flag
- Verified against the target cluster's live state that after this merge no Kustomization is left
  at `prune: true` except the prometheus pair

## Gate before the next stage

```
kubectl get kustomizations.kustomize.toolkit.fluxcd.io -A \
  -o custom-columns='NAME:.metadata.name,PRUNE:.spec.prune' \
  | grep -i true | grep -vE 'prometheus'
# must return nothing
```

Note the reconciler reads `main` on a 1-minute interval, so this is live shortly after merge.

### DIFF
--- a/flux/ack-build-cluster.yaml
+++ b/flux/ack-build-cluster.yaml
@@ -20,7 +20,10 @@ spec:
     kind: GitRepository
     name: test-infra
   path: ./flux/ack/build-cluster
-  prune: true
+  # prune deliberately disabled for the Argo CD migration: converting this path
+  # to a chart makes every raw object stale, and prune deletes stale objects.
+  # Must be live before any content change (see docs, D19-P1).
+  prune: false
   deletionPolicy: WaitForTermination
   targetNamespace: ack-system
   postBuild:
--- a/flux/ack.yaml
+++ b/flux/ack.yaml
@@ -25,7 +25,10 @@ spec:
     kind: GitRepository
     name: test-infra
   path: ./flux/ack/capability/role
-  prune: true
+  # prune deliberately disabled for the Argo CD migration: converting this path
+  # to a chart makes every raw object stale, and prune deletes stale objects.
+  # Must be live before any content change (see docs, D19-P1).
+  prune: false
   deletionPolicy: WaitForTermination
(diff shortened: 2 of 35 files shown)


knottnt (approved):
/lgtm

PR #1060, opened 2026-08-20
Render the ACK and Prow paths from Helm charts (stage 2/8)

gustavodiaz7722:
**Stage 2 of 8** in the Flux → Argo CD migration. See `docs/argocd-migration-runbook.md`.

Argo CD has no equivalent of Flux's `postBuild` substitution, so every path carrying a `${TOKEN}`
has to become a Helm chart before it can be handed over. This stage does that conversion **while
Flux is still the reconciler**, so it can be verified in isolation.

## Why this is safe to merge on its own

The Kustomization paths are unchanged. Each directory's contents become a `HelmRelease` pointing
at a new chart, and the HelmRelease reads the same values from the same `self-managed-vars`
ConfigMap via `valuesFrom`. helm-controller renders what kustomize used to render. Nothing here
requires Argo CD to exist.

## Verification

Rendered both sides and compared object sets per path — `kustomize build` at the previous revision
with the substitution applied, against `helm template` of the new chart with the values its
`HelmRelease` supplies:

| paths compared | result |
|---|---|
| 7 of 9 | identical object set |
| `flux/ack/cluster/addons` | `external-dns-role` no longer rendered here — see below |
| `flux/ack/cluster` | hub `Cluster` CR dropped — **intentional**, see below |

**`external-dns-role`** moved out of the `addons` path because the old `kustomize build` recursed
into `addons/roles/`. That subdirectory has its own Kustomization (`ack-addons-roles`) which renders
it identically, so no object loses an owner — the ownership boundary just got cleaner.

**The hub `Cluster` CR is deliberately dropped.** It carried the last render-time cluster read in
the repo, which Argo CD cannot satisfy: it renders off-cluster, so Helm's `lookup` returns empty. No
later stage re-adopts it — in the final state only `ack-build-infra` declares a `Cluster`, and that
one is the build cluster's. The live CR is therefore declared nowhere from here on, which is safe
because:

- prune is already disabled everywhere (stage 1)
- the CR carries `services.k8s.aws/deletion-policy: retain`, so even deleting it leaves the EKS
  cluster alone
- Terraform owns the actual cluster via `aws_eks_cluster.this` — the CR was only ever an adoption

## Gate before the next stage

All unsuspended Kustomizations `Ready=True`, and Prow's Deployment uids unchanged. A conversion that
were not inert would show up as recreates, and Prow's Deployments carry an immutable
`spec.selector`, so a recreate would take the control plane down.

```
kubectl -n prow get deploy -o json | python3 -c \
  'import json,sys; print(sorted((d["metadata"]["name"],d["metadata"]["uid"]) for d in json.load(sys.stdin)["items"]))'
```

Note the reconciler reads `main` on a 1-minute interval, so this is live shortly after merge.


knottnt (approved):
/lgtm

PR #1061, opened 2026-08-20
Stand up the Argo CD capability without cluster-admin (stage 3/8)

gustavodiaz7722:
**Stage 3 of 8** in the Flux → Argo CD migration.

Stands up the Argo CD capability and authorises it **without granting cluster-admin**. Flux still
owns every object; no Applications exist yet, so everything added here is inert until stage 4.

**Terraform only — 13 files, all under `bootstrap/`.** No Flux manifests, no chart changes, nothing
suspended. The path conversions landed separately in #1060 and #1062.

## What this adds

| file | what |
|---|---|
| `argocd.tf`, `capability.tf` | the capability itself, and the Identity Center lookup it consumes |
| `argocd-access.tf` | EKS access policy associations, hub cluster registration, the AppProject |
| `argocd-rbac.tf` | one narrow ClusterRole for the four cluster-scoped kinds no access policy covers |
| `argocd-logs.tf` | CloudWatch vended log delivery for the controllers |
| `identity/` | **a separate Terraform stack** — the IdC instance and the `<stack>-argocd-admins` group |
| `images.tf`, `flux.tf`, `variables.tf` | gate two one-shots behind variables defaulting `false` |

## Apply order, and the step that is easy to miss

**`bootstrap/identity/` must be applied first, and it keeps its own state** — deliberately, so the
main stack's destroy/apply cycle can never delete the IdC instance. It therefore needs its own
generated `backend.tf`. Skip it and the main stack fails *at plan*, not at apply:

> `Error: Missing required argument` — `data.aws_identitystore_group.argocd_admins`,
> `The argument "identity_store_id" is required, but no definition was found.`

which is `one(data.aws_ssoadmin_instances.this.identity_store_ids)` returning `null` in an account
with no instance.

Then, because `kubernetes_manifest` validates against the live API at plan time, the capability must
exist before the rest can plan:

```
cd bootstrap/identity && terraform init && terraform apply     # separate state
cd ..                 && terraform apply -target='awscc_eks_capability.argocd'
                         terraform apply                        # the rest
```

Run prod without `-auto-approve` and read the plan. Expect two `null_resource` destroys —
`bootstrap_prow_images` and `bootstrap_prow_images_job` dropping to `count = 0`. Both are inert:
neither carries a destroy-time provisioner.

## Why the authorisation looks like this

The capability's auto-created access entry has `kubernetesGroups: []` and a session-templated
username, so there is no stable RBAC subject — the `eks-access-entry:<arn>` group the AWS docs suggest
does not exist and binds nothing. The levers are which access *policies* are associated, plus adding a
Kubernetes group to the entry so a narrow ClusterRole can be bound for `StorageClass`, `IngressClass`,
`NodePool` and `NodeClass`. The only single policy covering those is `AmazonEKSClusterAdminPolicy`,
i.e. cluster-admin — which is what this avoids. The files record the alternatives that were tried and
failed, so they are not retried.

## Gate before stage 4

Grants must be live before any Application syncs. A sync attempted first fails on escalation
prevention with a message naming the *chart's* ClusterRole rather than the missing grantor rule, which
sends you to the wrong file.

```
for r in storageclasses ingressclasses nodepools.karpenter.sh namespaces clusterroles; do
  echo "$r: $(kubectl auth can-i create $r -A --as=probe --as-group=argocd-cluster-scoped)"
done
kubectl auth can-i delete namespaces -A --as=probe --as-group=argocd-cluster-scoped   # must be no
```

`can-i` reports `no` for anything granted by an EKS access policy — those are enforced by the EKS
authorizer and invisible to a SubjectAccessReview. Use it only for the in-cluster rules.

## Verification

- `terraform validate` passes
- rebased on `main` @ `7ce1aea`; merge-base diff is exactly the 13 files above
- commit set matches the runbook's Stage 3 assignment exactly
- no `suspend:` changes, no `docs/` changes

### DIFF
--- a/bootstrap/argocd-access.tf
+++ b/bootstrap/argocd-access.tf
@@ -1,0 +1,358 @@
+################################################################################
+# Argo CD authorization and cluster registration
+#
+# Everything here is pure Terraform. No local-exec, no kubectl.
+#
+# What EKS already did for us when the capability was created (verified live):
+#   AmazonEKSArgoCDClusterPolicy  scope=cluster
+#   AmazonEKSArgoCDPolicy         scope=namespace, namespaces=[argocd]
+#
+# Those cover Argo CD's OWN operation - reading its CRs, reading cluster
+# registration Secrets, API discovery, namespace creation, CRD create. They do NOT
+# grant permission to deploy workloads or to read arbitrary resources for health
+# assessment. That is what this file adds.
+#
+# Authorization is granted via EKS access policies rather than in-cluster RBAC,
+# because Argo CD cannot apply the object that authorizes Argo CD. The repo already
+# documents this pattern for Flux in flux/ack/build-cluster/access-entries.yaml.
+################################################################################
+
+################################################################################
+# SCOPE: hub cluster only.
+#
+# Nothing here touches the build cluster, deliberately. The build cluster is not
+# registered as an Argo CD spoke until Phase 4, and when it is, its AccessEntry
+# belongs in flux/ack/build-cluster/access-entries.yaml as an ACK CR alongside the
+# eight already there - not in Terraform. See docs/argocd-migration.md, D13.
+################################################################################
+
+locals {
+  argocd_capability_role_arn = aws_iam_role.argocd_capability.arn
+
+  # Namespaces Argo CD may write to on the hub.
+  #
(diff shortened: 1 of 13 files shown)


knottnt (approved):
/lgtm

PR #1064, opened 2026-08-20
Render the Applications from git via a root Application (stage 4/8)

gustavodiaz7722:
**Stage 4 of 8** in the Flux → Argo CD migration.

Moves the Argo CD Application definitions into git behind a single root Application, and gives the
Prow namespaces and ServiceAccounts an owner.

## ⚠️ This is the merge where Argo CD starts acting on live objects

Not stage 5, despite what the sequence suggests. The Applications are rendered with
`automated: true`, so as soon as the root Application exists they sync and begin **adopting objects
Flux is still reconciling.**

That overlap is intended. Adoption happens under `ServerSideApply=true`, so Argo CD takes over field
ownership *in place* rather than recreating — which matters most for Prow's Deployments, whose
immutable `spec.selector` makes a recreate an outage rather than a blip. Both reconcilers are live
from here until stage 5 suspends the Flux side.

**Apply the Terraform promptly after merging.** This PR adds the `namespaces` grant that wave 0's
`prow-namespaces` Application needs. Until it lands, wave 0 cannot sync, and because the root waits
for each wave to report healthy, every later wave sits `OutOfSync` with no error of its own — a stall
rather than a permission error. Nothing is damaged: those Namespaces and ServiceAccounts already
exist and this path adopts them.

## What this adds

| file | what |
|---|---|
| `argocd/applications/` | the app-of-apps chart: one template rendering 20 children from `values.yaml` |
| `bootstrap/argocd-applications.tf` | one `kubernetes_manifest` for the root Application, and the 16 values threaded into it |
| `prow/namespaces/` | the 2 Namespaces and 5 ServiceAccounts, moved out of `flux/ack/cluster/pod-identities/` |
| `bootstrap/argocd-rbac.tf` | the `namespaces` get/list/watch/create/update/patch grant — **no delete** |

Paths move out of Terraform and into git, so changing one is a commit rather than a `terraform apply`.

## Why `prow/namespaces` exists

The conversion left seven objects with no owner. Flux applied `prow-namespaces.yaml` directly from
`flux/ack/cluster/pod-identities/`, alongside the chart rather than inside it, because no chart may own
a Namespace — a Helm uninstall or rollback targets the namespace and cascades to everything in it. So
the `ack-pod-identities` Application, which points at the chart, never covered them. They survived only
because prune is disabled, and a fresh bootstrap would not have created them at all.

Everything targeting `prow` or `test-pods` therefore depends on this, which is why it is wave 0 with
`CreateNamespace=false` everywhere.

## Verification

- 8 files; no `suspend:` changes, no `docs/` changes
- commit set matches the runbook's Stage 4 assignment exactly
- `terraform validate` passes
- chart renders 20 children across waves 0–5; **every Application `path` verified to exist in the tree**
- `namespaces` grant present, and it excludes `delete`

## Gate before stage 5

Every Application Synced and Healthy, and every adopted object's uid held — compare against the stage 0
baseline. The paths where a recreate is destructive are `prow-config` (nine Deployments with immutable
`spec.selector`), `prow-namespaces` (namespaces cascade), and `prow-crds` (the ProwJob CRD and every
ProwJob with it).

Diagnose the **lowest** unhealthy wave; anything above it was never attempted.

```
kubectl -n argocd get applications.argoproj.io \
  -o custom-columns='WAVE:.metadata.annotations.argocd\.argoproj\.io/sync-wave,NAME:.metadata.name,SYNC:.status.sync.status,HEALTH:.status.health.status' \
  --sort-by='.metadata.annotations.argocd\.argoproj\.io/sync-wave'
```

`OutOfSync` on nothing but Flux's `kustomize.toolkit.fluxcd.io/*` labels is expected here — content
matches, ownership metadata does not.


knottnt (approved):
/lgtm

PR #1067, opened 2026-08-21
[test-infra] Cut every path over to Argo CD (stage 5/8)

gustavodiaz7722:
Stage 5 of the Flux to Argo CD migration. Follows #1064 and #1066.

## What this does

Suspends every converted path's HelmRelease or Kustomization in git, so Flux stops reconciling objects Argo CD already owns.

**It does not enable `automated` sync — that happened in Stage 4**, where the Applications are rendered with `automated: true` from the start. So the transition is the reverse of how it reads: Stage 4 is where Argo CD started acting, adopting objects under `ServerSideApply=true` while Flux was still reconciling them. This is where Flux stops. The window with both reconcilers live was #1064 to here.

Eight suspends, split across the two Kustomizations that deliver them:

| Delivered by | Suspends |
|---|---|
| `test-infra` (`./flux`) | `prow-crds`, `prow-build-cluster-resources`, `secrets` |
| `prow-charts` (`./flux/prow/charts`) | `prow-config`, `prow-data-plane`, `prow-agent-workflows`, `prow-jobs`, `prow-plugins` |

Suspend rather than delete throughout: deleting a Kustomization is what triggers garbage collection, whereas suspending stops reconciliation and pruning both and reverses by flipping one field.

## Also in this commit

- **`prow-build-cluster-resources` gets `automated`.** Its Application is composed at runtime by the connection Job rather than rendered from the chart, and it had no sync policy, so suspending the Flux Kustomization without this would leave those 13 build-cluster objects reconciled by nothing. Both halves are in this one commit for that reason. The edit reaches the live Application because it is drift on a *tracked* ConfigMap, so Argo CD syncs it, which re-runs the PostSync Job, which applies the updated Application.
- **`ServerSideApply=true` as a per-resource annotation on the ProwJob CRD.** The manifest is 667 KB and client-side apply stores it in `last-applied-configuration`, which the API server rejects at 262144 bytes. The Application-level sync option was not enough — Argo CD applied it client-side anyway through a manual sync, an automated sync, and with `Replace=true`. The annotation travels with the manifest.
- **Connection Job memory limit to 512Mi.** 128Mi was OOMKilled three times, surfacing only as a PostSync hook stuck in `Running`. The script's steps are `kubectl` pipelines, so peak is several concurrent Go binaries, not sequential.
- Comment-only fixes replacing references to `flux/argocd/`, which no longer exists, with `bootstrap/argocd-rbac.tf`.

## Pre-merge state on prod

The runbook's gate for this stage is that every Kustomization delivering a suspend is at current `main`, since a stale one silently swallows the change:

- `test-infra` and `prow-charts` both at `ceb02f5c`, `Ready=True`.
- 24 of 25 Kustomizations at `ceb02f5c`; the exception is `prow-build-cluster-connection`, which cannot render under Flux by design.
- 7 of the 8 suspended paths' Applications are `Synced/Healthy` with `automated`. The 8th is `prow-build-cluster-resources`, which this commit fixes.

## Verification

- `kustomize build ./flux` → 21 objects, 3 suspended; `./flux/prow/charts` → 5 objects, 5 suspended. Eight total, matching the table.
- Both modified Helm charts render; rendered Application carries `automated: {prune: false, selfHeal: false}` and the Job carries a 512Mi limit.
- `terraform validate` passes; no Terraform change in this stage, so **no `terraform apply` needed**.
- `flux/kustomization.yaml` is unchanged from `main`: the staged commit also removed `argocd.yaml` there, which #1065 already did, and dropped `prometheus.yaml`, which must stay until Stage 6.

## Post-merge

Gate is that nothing unsuspended is unhealthy and every Application is automated, then exercise it end to end by triggering a presubmit and confirming it schedules onto the build cluster.

Expect Flux to briefly report a batch of `DependencyNotReady` failures right after merge. That is it re-walking its graph as the revision propagates and it drains in about two minutes.

### DIFF
--- a/flux/ack/charts/ack-build-infra/templates/access-entries.yaml
+++ b/flux/ack/charts/ack-build-infra/templates/access-entries.yaml
@@ -162,12 +162,13 @@ spec:
 # WHY ClusterAdminPolicy, when the hub deliberately avoids it.
 #
 # On the hub, cluster-scoped access is granted by adding a kubernetesGroups entry to the
-# access entry and binding a narrow ClusterRole (flux/argocd/cluster-scoped-rbac.yaml),
-# specifically to avoid handing Argo CD cluster-admin. That approach cannot work here.
+# access entry and binding a narrow ClusterRole (bootstrap/argocd-rbac.tf), specifically
+# to avoid handing Argo CD cluster-admin. That approach cannot work here.
 # The ClusterRole would have to live inside THIS cluster, and once Flux is gone the only
 # thing able to apply it is Argo CD - the very thing it authorises. On the hub that
-# circularity is escaped because Flux owns flux/argocd/ until Phase 5 decides otherwise;
-# on a remote cluster there is no such applier. The narrow role would also have to
+# circularity is escaped because Terraform owns those objects and reaches the cluster
+# directly; on a remote cluster there is no such applier - and Terraform must hold nothing
+# keyed to a cluster ACK owns (D13), so it cannot be the applier here. The narrow role would also have to
 # enumerate every permission in those four Roles, or Kubernetes escalation prevention
 # refuses them: a principal may not create a Role granting permissions it lacks.
 #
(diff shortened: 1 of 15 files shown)


knottnt (approved):
/lgtm

PR #1069, opened 2026-08-21
[test-infra] Remove kube-prometheus-stack and the Prow Grafana dashboards (stage 6/8)

gustavodiaz7722:
Stage 6 of the Flux to Argo CD migration. Independent of the other stages — nothing in Argo CD replaces these, so they are removed outright.

## What this does

Removes `kube-prometheus-stack` and the four Prow Grafana dashboards from git: `flux/prometheus.yaml`, `flux/prometheus/`, and `prow/prometheus-dashboards/`. Almost all of the 5721 deleted lines are dashboard JSON.

Also drops the two rows in `flux/README.md` that documented the deleted files, and updates a comment in `flux/prow.yaml` that referred to these paths.

## Cleanup after merge

`prune: false` on the parent Kustomization means removing the declarations does not delete anything, so the live objects are collected by hand:

| # | delete |
|---|---|
| 1 | `prometheus` and `prometheus-dashboards` Kustomization objects |
| 2 | `prometheus` HelmRelease |
| 3 | the `prometheus` namespace |
| 4 | 10 `monitoring.coreos.com` CRDs |
| 5 | `prometheus-prometheus-kube-admission` Secret |

Deleting the `prometheus` Kustomization first makes step 2 cascade: its finalizer collects its inventory, which uninstalls the Helm release and most of the stack's objects with it.

When removing the CRDs, match on the `monitoring.coreos.com` API group rather than the name. `alertmanagerdefinitions`, `loggingconfigurations`, `rulegroupsnamespaces` and `workspaces` in `prometheusservice.services.k8s.aws` belong to ACK's Amazon Managed Prometheus controller and stay.

## Verification

- `kustomize build ./flux` goes from 21 objects to 19, the two Kustomizations removed, with no prometheus object left in the built tree.
- All 20 Argo CD Applications still resolve; none points at a deleted path.
- `terraform validate` passes. No Terraform change, so no `terraform apply`.


knottnt (approved):
/lgtm

PR #1071, opened 2026-08-21
[test-infra] Remove Flux (stage 7/8)

gustavodiaz7722:
Stage 7 of the Flux to Argo CD migration, second of two PRs. Follows #1070, which moved the build-cluster connection chart into `ack-system` so `flux-system` can go.

Three commits, in dependency order:

- `feat(argocd): Remove Flux` — deletes every Kustomization and HelmRelease definition, plus the raw manifest directories the charts replaced.
- `chore(bootstrap): Remove Terraform's Flux footprint` — `flux.tf`, `bootstrap-flux.sh`, `validate-kustomizations.sh`, the node role's `repository/fluxcd/*` ECR pull-through-cache policy, and `var.flux_version`.
- `refactor(argocd): Drop flux-system from the hub write grant` — withdraws the grant now that nothing renders into that namespace.

## What stays

The charts under `flux/` are Argo CD Application sources and are untouched: `flux/ack/charts/`, `flux/prow/charts/`, `flux/prow/crds/` and `flux/secrets/`. Only the Kustomization and HelmRelease *definitions* are deleted. The directory name is now a misnomer; renaming it means touching every Application path, so it is a separate change.

The vendored `flux2` chart under `charts/` and the 74 `kustomize.toolkit.fluxcd.io/prune: disabled` annotations go in Stage 8, along with `bootstrap/README.md`, which still documents the Flux workflow.

`ack-flux` is removed from the app-of-apps, taking the Application count from 20 to 19.

## Verification

- All 19 Application paths still resolve; app-of-apps renders 19.
- The only remaining `kind: HelmRelease` in the tree is inside the vendored `flux2` chart.
- `self-managed-vars` now appears only in comments — `bootstrap-prow-images.sh` takes those values from the provisioner instead, so a fresh bootstrap no longer depends on a ConfigMap that goes with the namespace.
- `terraform validate` passes.

## Applying

`terraform plan` is 1 add, 2 change, 9 destroy. Two things worth knowing before the apply:

The four `null_resource`s in `flux.tf` — `flux_system_namespace`, `bootstrap_flux`, `validate_kustomizations`, `restore_flux_registry` — carry create-time provisioners only, no `when = destroy`, so removing them runs nothing.

The apply removes the build cluster from the AppProject's destinations. That entry is added at runtime by the connection Job, because D13 keeps Terraform from naming a cluster ACK owns, so Terraform plans to drop a field it does not declare. `prow-build-cluster-resources` stops syncing until the Job re-patches it, so the apply is followed by a refresh of `prow-build-cluster-connection` to re-run its PostSync Job. This is not specific to these commits — any apply does it — and the durable fix is a separate change.

Cluster steps in the runbook order after merge: scale the Flux controllers to zero, clear finalizers, delete the Flux CRs and CRDs, cluster RBAC and the `flux-system` namespace, then delete Flux's ACK CRs now that git no longer declares them, then apply Terraform.


knottnt (approved):
/lgtm

PR #1073, opened 2026-08-21
[test-infra] Sweep the remaining Flux references and remove the nodepool swap (stage 8/8)

gustavodiaz7722:
Stage 8, the last of the Flux to Argo CD migration. Two commits, both cleanup of things that were only load-bearing while Flux existed. Nothing here blocks anything and it is safe to defer.

## The sweep

Deletes the vendored `flux2` chart (46 files under `charts/`) and `scripts/pull-flux-chart.sh` that fetched it, and rewrites the comments and docs that described Flux mechanisms into their Argo CD equivalents.

The part to notice is **`scripts/upgrade-prow.sh`**. It read the Prow version from the `prow-version` ConfigMap in `flux-system`, so it has been broken since that namespace was deleted in stage 7 and would have failed the next time anyone bumped Prow. It now reads and writes the three chart values files directly. If there is automation or a documented procedure that calls it, re-read it after merging — both its inputs and the files it edits changed.

`bootstrap/README.md` also loses its "Vendor the Flux chart" bootstrap step and its "Upgrading Flux" section, the latter replaced with an "Upgrading Argo CD" note explaining that the control plane is an EKS capability managed off-cluster, so there is no chart to vendor and nothing here pins its version.

Two leftovers I removed beyond the staged commit, both in `bootstrap/README.md`:

- The `ecr-pullthroughcache/ghcr-fluxcd` secret and its resource policy were still in the first-time bootstrap steps, annotated with a note saying they are no longer required. A bootstrap instruction that tells you not to follow it is worse than its absence — the cache rule it served went with Flux, so the block is gone and replaced by one line recording that.
- The `bootstrap-env.sh` walkthrough still listed "flux version" among the variables it prompts for. `var.flux_version` was removed in stage 7.

## The nodepool swap

Removes `null_resource.swap_nodepool` and `bootstrap/scripts/swap-nodepool.sh`, which deleted the built-in `general-purpose` NodePool once `prow-compute` was Ready. It existed because Flux ran on-cluster and needed capacity before the real pools were created; Argo CD and ACK both run off-cluster, so nothing needs capacity before the first sync.

It was also never durable, which is the better reason to drop it. The NodePool carries `app.kubernetes.io/managed-by: eks` with no owner references or field managers, so while `compute_config.node_pools` lists it the EKS control plane reconciles it back after deletion. The `null_resource` was fighting the control plane and losing.

This also drops the last `depends_on` edge into `null_resource.validate_kustomizations`, which polled Flux Kustomizations and has nothing left to poll.

## Deliberately not included

The 75 `kustomize.toolkit.fluxcd.io/prune: disabled` annotations on chart-rendered objects stay. They are inert — the controller that read them is gone — but removing them is a mechanical change across many chart templates and not a no-op at runtime: the annotation is rendered onto every ACK CR, so dropping it from the charts makes Argo CD strip it from live objects on the next sync. Metadata-only, but worth doing deliberately rather than inside this stage.

## Verification

- 80 files, −10942 lines, almost all of it the vendored chart.
- `charts/` and `scripts/pull-flux-chart.sh` are gone; no reference to either remains.
- App-of-apps still renders 19 Applications and every Application path resolves.
- `terraform validate` passes.
- No Terraform change beyond removing `null_resource.swap_nodepool`, so the apply is a single destroy of a resource whose script is gone.


knottnt (approved):
/lgtm

PR #1093, opened 2026-09-08
Upgrade Prow to v20260908-104d452f4 and bump patch revision

gustavodiaz7722:
## Description

Upgrades the Prow control plane to `v20260908-104d452f4` (from `v20260702-fff8399c7`) and tools to `v20260729-f66136c16c` (from `v20260515-778139c32d`), keeping us current with upstream. Bumps `prowPatchRevision` to `2`.

Motivated by outdated OS packages in the currently deployed control-plane images. Vulnerability scanning flagged `curl`, `expat` and `openssl` across every component running at `v20260702-fff8399c7-ack.1` — `crier`, `deck`, `ghproxy`, `hook`, `horologium`, `prow-controller-manager`, `sidecar`, `sinker`, `status-reconciler` and `tide`.

The patch revision bump is the operative part of the fix. As `prow-mirror`'s own comment explains, upstream pins its ko base images in `.ko.yaml` and refreshes that pin only every few months, so a Prow release ships outdated `expat`/`openssl`/`curl` **regardless of how recent the tag is**. The `apk upgrade` during mirroring is what patches them. Because `-ack.2` does not yet exist in ECR, the mirror's existence check will not skip it and all 15 mirrored images get rebuilt against current OS packages.

Note that the version bump alone would not have been sufficient — a newer upstream tag still carries the stale base. The revision bump is what forces the rebuild.

## Changes

- `prow/config/values.yaml` — `imageMirror.prowVersion`, `imageMirror.prowPatchRevision`
- `prow/jobs/values.yaml` — `prowVersion`, `toolsVersion`, `prowPatchRevision`
- `flux/prow/charts/prow-mirror/values.yaml` — same three
- `prow/config/Chart.yaml` — 0.4.24 → 0.4.25
- `flux/prow/crds/prowjob_customresourcedefinition.yaml` — refreshed from upstream

All three version copies move together, as `scripts/upgrade-prow.sh` requires.

## CRD upgrade

The refresh moves the CRD to controller-gen `v0.22.0`. Verified additive:

- 22 top-level `spec` properties before and after, none removed
- Single `v1` version, served and stored — no migration
- `preserveUnknownFields: false` retained, so the existing `ignoreDifferences` entry in `argocd/applications/values.yaml` stays valid
- New nested fields (`bindMountOptions`, podCertificate volume source, container restart rules, resize policy) are optional pod-spec additions from a newer Kubernetes API
- The apparent `mountPath` removal in the diff is `bindMountOptions` sorting ahead of it; the count is 3 both before and after

**Re-added the `argocd.argoproj.io/sync-options: ServerSideApply=true` annotation**, which `upgrade-prow.sh` strips when it overwrites the file from upstream. Without it Argo CD applies the CRD client-side and the API server rejects the `last-applied-configuration` annotation as too long. The manifest grew 667 KB → 704 KB, so the 262144-byte limit is further out of reach than before, and the comment's stated size was updated to match.

## Testing

- `helm template` on `prow/config` — all 9 components and 4 utility images compose to `v20260908-104d452f4-ack.2`
- `helm template` on `prow/jobs` — `PROW_VERSION`, `TOOLS_VERSION` correct, `PATCH_SUFFIX: "-ack.2"` renders as a string
- `helm template` on `prow-mirror` — targets both new upstream tags
- `kubectl kustomize flux/prow/crds` — builds clean, sync-options annotation present in output
- No references to the previous versions remain anywhere in the repo

## Follow-ups (not in this PR)

- A container is running `ack-prow-webhook:v1.0.30`, an image name that appears nowhere in this repo — we build `agent-plugin` per `prow/plugins/images_config.yaml`. Looks like an orphaned Deployment running a pre-rename artifact, and its packages are stale. Needs a cluster-side look rather than a repo change.
- `scripts/upgrade-prow.sh` has two rough edges worth fixing: `yq -i` reformats the values files (strips all blank lines), and the CRD `cp` silently drops the sync-options annotation.

### DIFF
--- a/flux/prow/charts/prow-mirror/values.yaml
+++ b/flux/prow/charts/prow-mirror/values.yaml
@@ -14,6 +14,6 @@ region: ""
 #
 # PROW_PATCH_REVISION is a monotonic counter appended to mirrored image tags and must
 # never be reset. PROW_VERSION and TOOLS_VERSION move independently but share it.
-prowVersion: "v20260702-fff8399c7"
-toolsVersion: "v20260515-778139c32d"
-prowPatchRevision: "1"
+prowVersion: "v20260908-104d452f4"
+toolsVersion: "v20260729-f66136c16c"
+prowPatchRevision: "2"
--- a/flux/prow/crds/prowjob_customresourcedefinition.yaml
+++ b/flux/prow/crds/prowjob_customresourcedefinition.yaml
@@ -4,12 +4,12 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     api-approved.kubernetes.io: https://github.com/kubernetes/test-infra/pull/8669
-    controller-gen.kubebuilder.io/version: v0.17.3
+    controller-gen.kubebuilder.io/version: v0.22.0
     # THE ONLY LOCAL EDIT TO THIS UPSTREAM FILE. scripts/upgrade-prow.sh regenerates it from
     # kubernetes/test-infra, so this line has to be re-added after an upgrade or Argo CD stops
     # being able to apply the CRD at all.
     #
-    # This manifest is 667 KB. Client-side apply stores the whole thing in
+    # This manifest is 704 KB. Client-side apply stores the whole thing in
     # kubectl.kubernetes.io/last-applied-configuration, which the API server rejects:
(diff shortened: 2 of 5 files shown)


knottnt (approved):
/lgtm

PR #1110, opened 2026-09-15
Make Argo CD sync waves actually order the app-of-apps rollout

gustavodiaz7722:
## Description

The `argocd.argoproj.io/sync-wave` annotations on the 19 child Applications have never had any effect. Two independent causes, both fixed here.

**Nothing could be waited on.** Argo CD removed the built-in health assessment for `argoproj.io/Application` in 1.8 ([argo-cd#3781](https://github.com/argoproj/argo-cd/issues/3781)), and [its docs](https://argo-cd.readthedocs.io/en/latest/operator-manual/health/) name this pattern as the case that has to restore it. Nothing did, so every child reported no health to `root-applications`:

```
kubectl -n argocd get application root-applications -o jsonpath='{.status.resources[*].health}'
-> empty, for all 19
```

The [EKS capability docs](https://docs.aws.amazon.com/eks/latest/userguide/argocd-configure-settings.html) state the consequence directly: a resource with no health check is excluded from the parent's health, and *"sync waves can advance before those resources are ready, because sync ordering depends on reported health."*

**And the parent was not the thing syncing.** Children tracked the same branch as the root, so a commit changing only chart *contents* rendered byte-identical child specs. The root had no drift and never synced, while each child noticed its own source had changed and synced itself, unordered. Waves are evaluated inside a sync operation, and for an app-of-apps that means the root's — so none was ever processed. On the upgrade that broke `prow-config`, the root's last sync was 18 days earlier.

Together those put wave 4 after wave 5, in both directions:

```
bootstrap  prow-config (5) 00:45:24   prow-mirror (4) 00:45:41
upgrade    prow-config (5) 21:46:58   prow-mirror (4) 21:54:48
```

`prow-config` pins the image tags `prow-mirror` creates, so the nine control-plane Deployments rolled onto tags that did not exist yet: 11 minutes of ImagePullBackOff, with `ghproxy`, `horologium` and `tide` fully down because they are `strategy: Recreate`.

## Changes

**Pin children to the resolved commit SHA** (`bootstrap/argocd-applications.tf`, `argocd/applications/`). Carried from the root as a helm parameter holding `$ARGOCD_APP_REVISION`; build-env interpolation only happens in `parameters`, so it cannot travel in the values blob. Every commit now changes all 19 child specs, so the root always syncs and always orders the rollout by wave, with a health wait at each step and a child that will not go Healthy stopping everything behind it.

Nothing writes a SHA to git — the root keeps tracking the branch, and Argo CD resolves it on every refresh. There were 20 independent moving refs racing; now there is one and 19 derived values.

**Add `argocd-cm` with the Application health check** (`bootstrap/argocd-config.tf`) so those waits are real. Terraform-owned for the reason the file already lists for the AppProject and the rest: a health check delivered by an Application would itself be applied by a wave, and it is what makes waves work. `root-applications` depends on it so ordering is right from the opening sync.

**Track the mirror Job** instead of running it as a Helm hook (`flux/prow/charts/prow-mirror/`). Hook objects are excluded from health assessment, so `prow-mirror` reported Healthy the moment its ConfigMap applied while the mirroring was still running. Tracked, Argo CD applies its Job health, so the path is Healthy only once the images are in ECR and Degraded if the mirror fails. Argo CD's `Replace=true` handles the immutable `spec.template`, as `prow/jobs` already relies on for `job-config-substitutor`.

**Force field-manager conflicts on the root.** Terraform owns this object's spec; without this a single out-of-band `kubectl patch` makes it permanently unmanageable.

## Verified on staging

Not inferred — watched end to end:

```
18:13:29  wave 2  ack-pod-identity-roles
18:13:36  wave 4  prow-mirror          <- Job starts, health goes Progressing
18:13:37  wave 4  prow-crds
18:13:57  wave 3  prow-build-cluster-connection
18:19:47  wave 5  prow-jobs            <- 6 min later, only after wave 4 went Healthy
18:19:49  wave 5  prow-config
```

Root sat at `waiting for healthy state of argoproj.io/Application/prow-crds` with wave 5 still on the previous revision throughout. Children came out pinned to a real 40-char SHA, confirming `$ARGOCD_APP_REVISION` interpolates. Final state: 19/19 children on the resolved SHA, root `Succeeded`.

## Three refinements staging forced

None were findable locally, and each is a case the health check has to get right.

- **No gate on `operationState.phase == "Running"`.** An operation can wedge indefinitely: `ack-cluster` sat Running for 27 days waiting on an `AccessEntry` that had been removed from its chart, while the app itself was Synced and Healthy. There is no clock in the Lua sandbox to age that out. The revision comparison is the real settling check.
- **OutOfSync holds the wave, except for pending prunes.** Prune is never rendered anywhere in this repo, so an app with an orphaned resource can never reach Synced — `prow-namespaces` held wave 0 over a `ServiceAccount` that had moved to another chart. Tolerated only when *every* out-of-sync resource is a pending prune; anything else still holds.
- **When a full SHA is pinned, `status.sync.revision` must match it.** A child carries Synced and Healthy from its previous revision for a moment after the parent repins it, and the parent would otherwise advance on that stale value. Gated on a 40-char target so a branch-pinned child cannot deadlock.

Ten cases covered including "orphan prune *and* a real change -> still holds".

## Merge ordering — please read

**`terraform apply` must land before this merges, per environment.** The chart requires `childRevision`, which the root supplies. Git-first breaks the root's render and stops every deploy. This matters because prod tracks `main` directly, so merging *is* deploying.

The root deliberately keeps passing `targetRevision` alongside the new parameter so the two halves roll forward or back in either order. Dropping it deadlocks both ways, which `terraform plan` caught before it reached a cluster.

## Behavioural change

Deploys serialize: every commit walks waves 0-5 with a health wait at each step, and one unhealthy app stops everything behind it. That is the intent — an ordered, fail-stop rollout — but it is a visible change from today, where the 19 children deploy independently.

## Follow-ups (not in this PR)

- `ack-cluster` has an operation Running since 2026-08-19 waiting on an `AccessEntry` that no longer exists in its chart. Needs terminating, and worth understanding how it vanished mid-sync.
- An orphaned `ServiceAccount/workflow-runner` in `test-pods` keeps `prow-namespaces` OutOfSync. Deleting it lets that path reach Synced.


sapphirew (approved):
/lgtm

